PR #7, opened 2019-08-20
Clinical Registration Endpoints

joneubank:
### Description of changes

Clinical Registration connections from the Clinical Service. Includes file upload, get current (without upload), clear, and commit.

### Type of Change

- [ ] Bug
- [ ] Styling
- [X] New Feature


schemas/Clinical/index.js
```diff
@@ -0,0 +1,204 @@
...
+    errors: get(data, 'errors', []).map((errorData, i) =>
+      convertRegistrationErrorToGql(errorData, i),
+    ),
+
+    newDonors: convertRegistrationStatsToGql(get(registration, 'stats.newDonorIds', {})),
+    newSpecimens: convertRegistrationStatsToGql(get(registration, 'stats.newSpecimenIds', {})),
+    newSamples: convertRegistrationStatsToGql(get(registration, 'stats.newSampleIds', {})),
+    alreadyRegistered: convertRegistrationStatsToGql(
+      get(registration, 'stats.alreadyRegistered', {}),
+    ),
```
hlminh2000: seems to be a lot of wrangling going on. I think it's possible make graphql lazy evaluate them by thunking like so (basically make them in-line resolvers): 
```
records: () => get(registration, 'records', []).map((record, i) =>
  convertRegistrationRecordToGql(record, i),
)
```
Give it a shot, if it works I'd recommend that to reduce server load :)

schemas/Clinical/index.js
```diff
@@ -0,0 +1,204 @@
...
+    Remove the Clinical Registration data currently uploaded and not committed
+    """
+    clearClinicalRegistration(shortName: String!, registrationId: String!): Boolean!
+
+    """
+    Complete registration of the currently uploaded Clinical Registration data
+    """
+    commitClinicalRegistration(shortName: String!, registrationId: String!): Boolean!
+  }
+`;
```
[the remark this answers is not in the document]
joneubank: Yes, thanks, will add those in.

schemas/Clinical/index.js
joneubank: Let me look into lazy resolving the fields. I'll update the PR...

hlminh2000: theoretically, that thunk is just a resolver so it receives the same arguments as the others as well, in case that's ever useful :)

joneubank:
@hlminh2000 The resolvers do a lot of calculation so I'll try to add field level resolvers... see whats doable.

My performance concern is with buffering the file upload before sending on to the Clinical Service... any thoughts with that?

services/clinical/index.js
```diff
@@ -0,0 +1,98 @@
+import { CLINICAL_SERVICE_ROOT } from '../../config';
+import fetch, { Response } from 'node-fetch';
+import { AuthenticationError, UserInputError, ServerError } from 'apollo-server-express';
+
+import FormData from 'form-data';
+
+/*
+convert the REST status codes to GQL errors, or return the response if passing
+*/
+const registrationResponseHandler = async response => {
```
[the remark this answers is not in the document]
ciaranschutte: cool cool we can make a more generic / turn this into something generic if need occurs

app.js
```diff
@@ -42,6 +43,7 @@ const init = async () => {
     context: ({ req }) => ({
       isUserRequest: true,
       egoToken: (req.headers.authorization || '').split('Bearer ').join(''),
+      Authorization: `Bearer ${(req.headers.authorization || '').split('Bearer ').join('')}` || '',
```
d8660091: Does this line do `Authorization: ${req.headers.authorization || ''}`

package.json
```diff
@@ -17,7 +17,8 @@
     "@babel/preset-env": "^7.4.2",
     "@babel/register": "^7.4.0",
     "@grpc/proto-loader": "^0.5.0",
-    "@icgc-argo/program-service-proto": "0.0.5",
+    "@icgc-argo/ego-token-utils": "^5.1.0",
+    "@icgc-argo/program-service-proto": "0.0.4",
```
d8660091: are we doing a downgrade?

joneubank: thanks, no, fixing

app.js
joneubank: almost... it ensures that the output is `Bearer {token}` whether the incoming header had the bearer prefix or not.

d8660091: For easier readable, I suggest `Bearer ${(req.headers.authorization || '').replace('Bearer ', '')`

joneubank: that is cleaner, will do

d8660091: FYI, I just realized trimStart could be used. https://lodash.com/docs/4.17.15#trimStart. That's not a big deal

joneubank: accomplished the same with regex ;)
`.replace(/^Bearer[\s]*/, '')`

PR #18, opened 2019-08-31
Gateway for File templates

KevinFHartmann:
Created a new endpoint for downloading template files called /download

It's not under /graphql because GraphQL only returns content-type JSON, but we need non-JSON content types (such as text/tab-separated-values) to download files.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [X] New Feature


app.js
```diff
@@ -58,6 +62,25 @@ const init = async () => {
...
+  // Our specification download service can't use GraphQL because GraphQL specification requires the content-type
+  // that it returns be json, and we want to be able to return other content types, such as tab-separated-values,
+  // so that the user is automatically prompted to save the file from their browser.
+  const apiRoot = urlJoin(CLINICAL_SERVICE_ROOT, SUBMISSION_TEMPLATE_PATH);
+  app.get('/downloads', async (req,res) => {
+    const name = req.query.template;
+    if (name == undefined) {
+      res.json({error: "Invalid query: use parameter template to specify the desired template file."});
+    } else {
+      const data = fetch(urlJoin(apiRoot, name), {synchronous: true}).then( (r) => {
```
joneubank: I couldn't find this `synchronous` option in the api docs, could you send me a reference for what it does and why we are using it?

app.js
```diff
@@ -58,6 +62,25 @@ const init = async () => {
     res.json(version);
   });
 
+  // Our specification download service can't use GraphQL because GraphQL specification requires the content-type
+  // that it returns be json, and we want to be able to return other content types, such as tab-separated-values,
+  // so that the user is automatically prompted to save the file from their browser.
+  const apiRoot = urlJoin(CLINICAL_SERVICE_ROOT, SUBMISSION_TEMPLATE_PATH);
+  app.get('/downloads', async (req,res) => {
```
joneubank: In order to keep route implementations separate from the main app/gql declaration, can we do the following:

- create a new directory called `routes`, with a file called `clinical.js`
- in `routes/clinical.js` lets create an [express.Router](http://expressjs.com/en/guide/routing.html#express-router) for clincialRoutes and define the implementation as `/template`. export this router.
- import the router in app.js and have it handle the path `/clinical`

End goal to resolve `/clinical/template` to the download code you wrote.

.env.schema
```diff
@@ -12,5 +12,6 @@ EGO_PUBLIC_KEY=
 PROGRAM_SERVICE_ROOT=
 
 CLINICAL_SERVICE_ROOT=
+SUBMISSION_TEMPLATE_PATH=
```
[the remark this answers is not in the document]
joneubank: Setting the default value in the config file (which you did) is a good approach as well. No change needed.

app.js
KevinFHartmann: It runs the fetch synchronously, and we're using it because I forgot to switch back to async once I got things working. ;-) Will fix.

routes/clinical.js
```diff
@@ -0,0 +1,24 @@
...
+    const data = fetch(urlJoin(apiRoot, name)).then((r) => {
+        res.status(r.status);
+        res.headers = r.headers;
+        // copying headers doesn't copy the content-type header...
+        res.contentType(r.headers.get('content-type'));
+        res.send(r.body.read());
+    });
+});
+
+module.exports = router
```
joneubank: New line

PR #22, opened 2019-09-03
Feature/upload clinical submission

jaserud:
**Description of changes**

Added schema and service for upload clinical submission. The logic is very similar to registration upload just for multiple files and clinicalTypes.


schemas/Clinical/index.js
```diff
@@ -76,6 +76,40 @@ const typeDefs = gql`
     donorId: String!
     specimenId: String!
   }
+  
+  type ClinicalSubmissionData @cost(complexity: 10) {
+    id: ID    
+    state: String
```
joneubank: If there is a known list of possible states, lets use an enum with those values.

schemas/Clinical/index.js
```diff
@@ -76,6 +76,40 @@ const typeDefs = gql`
...
+    version: String
```
joneubank: Are the versions integers?

schemas/Clinical/index.js
```diff
@@ -178,6 +220,59 @@ const convertRegistrationDataToGql = data => {
   };
 };
 
+const convertClinicalSubmissionDataToGql = (data) => {
+  const submission = get(data, "submission", {});
+  const schemaErrors = get(data, "errors", {});
+  // convert clinical entities for gql
+  const clinicalEntities = [];
+  for (var clinicalType in submission.clinicalEntities) {
```
joneubank: don't use `var` here, introduces clinicalType to the global scope. use `let` instead.

schemas/Clinical/index.js
```diff
@@ -178,6 +220,59 @@ const convertRegistrationDataToGql = data => {
...
+    clinicalEntities.push(convertClinicalSubmissionEntityToGql(clinicalType, submission.clinicalEntities[clinicalType]));
+  }
+  // collect schema errors for each entity in dataErrors (not sure if this is OK??)
+  for (var clinicalType in schemaErrors) {
```
joneubank: let instead of var

schemas/Clinical/index.js
```diff
@@ -178,6 +220,59 @@ const convertRegistrationDataToGql = data => {
...
+    console.log("Curent entity:" + clinicalType);
```
joneubank: remove console log

schemas/Clinical/index.js
```diff
@@ -235,6 +330,24 @@ const resolvers = {
...
+    uploadClinicalSubmissions: async (obj, args, context, info) => {
+      const { Authorization, egoToken } = context;
+      const { shortName, clinicalFiles } = args;      
+
+      // see reason in uploadRegistration
+      if (!TokenUtils.canWriteSomeProgramData(egoToken)) {
+        throw new AuthenticationError('User is not authorized to write data');
+      }
+
+      var filesMap = {};
```
joneubank: use const here

schemas/Clinical/index.js
jaserud: the version is a uuid

schemas/Clinical/index.js
jaserud: there is, I will add it then.

PR #37, opened 2019-09-16
add fileName and createdAt to Clinical resp + prettier

ciaranschutte:
**Description of changes**

add createdAt and fileName to clinical response 

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -12,10 +12,12 @@ const TokenUtils = createEgoUtils(EGO_PUBLIC_KEY);
 import costDirectiveTypeDef from '../costDirectiveTypeDef';
 import clinicalService from '../../services/clinical';
 import { ERROR_MESSAGES } from '../../services/clinical/messages';
+import customScalars from '../customScalars';
 
 const typeDefs = gql`
   ${costDirectiveTypeDef}
   scalar Upload
+  scalar DateTime
 
   enum SubmissionState {
     OPEN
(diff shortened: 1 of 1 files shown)


schemas/Clinical/index.js
```diff
@@ -214,6 +214,8 @@ const convertRegistrationDataToGql = data => {
   return {
     id: data._id || null,
     creator: data.creator || null,
+    fileName: data.batchName || null,
+    createdAt: data.createdAt || null,
```
hlminh2000: these are the only non-prettier changes right? if so, lgtm :)

hlminh2000:
actually, do we need to update the program-service protobuff?

schemas/Clinical/index.js
ciaranschutte: yup!

ciaranschutte:
@hlminh2000 needed to add extra fields to type

schemas/Clinical/index.js
```diff
@@ -31,7 +31,8 @@ const typeDefs = gql`
   type ClinicalRegistrationData @cost(complexity: 10) {
     id: ID
     creator: String
-
+    fileName: String
+    createdAt: String
```
hlminh2000: there's a `DateTime` scalar that should be used for date timestamps for consistency :)
See an example usage in: `schemas/Program/index.js`
Resolver for that scalar is in: `schemas/customScalars.js`

PR #38, opened 2019-09-16
Support tsv file

d8660091:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/routes/clinical.js
+++ b/routes/clinical.js
@@ -11,7 +11,7 @@ var router = express.Router();
 const apiRoot = urlJoin(CLINICAL_SERVICE_ROOT, SUBMISSION_TEMPLATE_PATH);
 
 router.get('/template/:template', async (req, res) => {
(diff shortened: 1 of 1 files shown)


joneubank:
What motivated this change? the template name should be provided without `.tsv`, if it isn't that feels like an oddity in the UI.  Do we intentionally want to support making `/template/registration` and `/template/registration.tsv`both valid paths?

d8660091:
@joneubank The default file name is [name] if we download a file from a link `/template/[name]`. If we use `/template/registration`. The file name would be `registration`. So to make the file name have a `.tsv` suffix. I made the change

joneubank:
Ok that makes sense. Thanks!

d8660091:
Thanks

PR #40, opened 2019-09-16
validate endpoint

jaserud:
added validate endpoint for clinical validate

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -154,6 +154,14 @@ const typeDefs = gql`
       shortName: String!,
       clinicalFiles: [Upload!]
     ): ClinicalSubmissionData! @cost(complexity: 30)
+    
+    """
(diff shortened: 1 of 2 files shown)


schemas/Clinical/index.js
```diff
@@ -354,6 +362,16 @@ const resolvers = {
       const response = await clinicalService.uploadClinicalSubmissionData(shortName, filesMap, Authorization);    
       return convertClinicalSubmissionDataToGql(response);
     },
+    validateClinicalSubmissions: async (obj, args, context, info) => {
+      const { Authorization, egoToken } = context;
+      const { shortName, version } = args;
+      // see reason in uploadRegistration
+      if (!TokenUtils.canWriteSomeProgramData(egoToken)) {
```
joneubank: Don't need this, not uploading files, clinical should provide this check as needed (it does)

PR #42, opened 2019-09-18
Feature/34 get submission

jaserud:
get submission from clinical


schemas/Clinical/index.js
```diff
@@ -281,6 +286,13 @@ const resolvers = {
       const response = await clinicalService.getRegistrationData(shortName, Authorization);
       return convertRegistrationDataToGql(response);
     },
+    clinicalSubmissions: async (obj, args, context, info) => {
+      const { Authorization, egoToken } = context;
```
joneubank: You don't use egoToken here, do you?

PR #47, opened 2019-09-19
🔊 adds graphql request logging query and variables

hlminh2000:
logs graphql query and variables for every gql request


app.js
```diff
@@ -23,6 +23,14 @@ ApolloServer.prototype._createGraphQLServerOptions =
 
 ApolloServer.prototype.createGraphQLServerOptions = async function(req, res) {
   const options = await this._createGraphQLServerOptions(req, res);
+  logger.info(
```
joneubank: this log is pretty verbose... maybe set it to debug level so we can filter out when not needed.

PR #48, opened 2019-09-23
Release Candidate 1.3.1

joneubank:
## New Features
- Get Active Clinical Submission

### DIFF
--- a/services/clinical/index.js
+++ b/services/clinical/index.js
@@ -63,6 +63,17 @@ const commitRegistrationData = async (programShortName, registrationId, Authoriz
   return response;
 };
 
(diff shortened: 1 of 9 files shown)


services/clinical/index.js
```diff
@@ -63,6 +63,17 @@ const commitRegistrationData = async (programShortName, registrationId, Authoriz
   return response;
 };
 
+const getClinicalSubmissionData = async (programShortName, Authorization) => {
+  const url = `${CLINICAL_SERVICE_ROOT}/submission/program/${programShortName}/clinical/upload`;
```
hlminh2000: just a note, generally would have suggested using `url-join` for url construction, but as long as it works :)

PR #50, opened 2019-09-23
Feat/#445 sample reg upload

ciaranschutte:
**Description of changes**

- Return error response for 400 http response eg. "Invalid file name"
- 422 http response return errors to client
- ClinicalRegistrationRecord change some fields to optional because not all fields have values
eg. if columns names are incorrect we don't have a 'value' field

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [ ] Matches design:

  - font size, weight, colour

- [x] Manual testing of UI feature.

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -376,7 +376,7 @@ const resolvers = {
 
         // Success data is inside the key "registration", error data is in the root level
         const data = { ...response.registration, errors: response.errors };
-        return convertRegistrationDataToGql(response);
+        return convertRegistrationDataToGql(data);
       } catch (e) {
         // errors that don't go into error table
         if (e.code) {
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -238,7 +238,6 @@ const convertRegistrationDataToGql = data => {
       get(data, 'records', []).map((record, i) => convertRegistrationRecordToGql(record, i)),
     errors: () =>
       get(data, 'errors', []).map((errorData, i) => convertRegistrationErrorToGql(errorData, i)),
-
     newDonors: () => convertRegistrationStatsToGql(get(data, 'stats.newDonorIds', {})),
     newSpecimens: () => convertRegistrationStatsToGql(get(data, 'stats.newSpecimenIds', {})),
     newSamples: () => convertRegistrationStatsToGql(get(data, 'stats.newSampleIds', {})),
@@ -367,15 +366,6 @@ const resolvers = {
       const { filename, createReadStream } = await registrationFile;
       const fileStream = createReadStream();
 
-      /**
-       *
-       *
-       *
-       *
-       *
-       *
-       *
-       */
       try {
         const response = await clinicalService.uploadRegistrationData(
           shortName,
@@ -383,7 +373,12 @@ const resolvers = {
           fileStream,
           Authorization,
         );
+
+        // Success data is inside the key "registration", error data is in the root level
+        const data = { ...response.registration, errors: response.errors };
+        return convertRegistrationDataToGql(response);
       } catch (e) {
+        // errors that don't go into error table
         if (e.code) {
           return { error: e.msg, code: ERROR_CODES[e.code] };
         } else {
@@ -391,12 +386,6 @@ const resolvers = {
           console.error('other error', e);
         }
       }
-
-      console.log('file name', filename, 'Resp', response);
-
-      // Success data is inside the key "registration", error data is in the root level
-      const data = response.successful ? response.registration : response;
-      return convertRegistrationDataToGql(data);
     },
     clearClinicalRegistration: async (obj, args, context, info) => {
       const { Authorization } = context;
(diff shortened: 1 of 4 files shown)


schemas/Clinical/index.js
```diff
@@ -300,16 +320,27 @@ const resolvers = {
...
+        const response = await clinicalService.uploadRegistrationData(
+          shortName,
+          filename,
+          fileStream,
+          Authorization,
+        );
+
+        // Success data is inside the key "registration", error data is in the root level
+        const data = { ...response.registration, errors: response.errors };
+        return convertRegistrationDataToGql(response);
```
ciaranschutte: we need to pass in errors too

schemas/Clinical/index.js
```diff
@@ -331,7 +331,7 @@ const resolvers = {
 
         // Success data is inside the key "registration", error data is in the root level
         const data = { ...response.registration, errors: response.errors };
-        return convertRegistrationDataToGql(response);
+        return convertRegistrationDataToGql(data);
```
ciaranschutte: we need to pass in errors too

joneubank: 👍

utils/restUtils.js
```diff
@@ -6,6 +6,7 @@ export const restErrorResponseHandler = async response => {
   switch (response.status) {
     case 200:
     case 201:
+    case 422:
```
joneubank: This looks good here but I am nervous that in the long run we'll have a 422 somewhere that shouldn't be ignored this way. Maybe you could handle this error in the specific service method instead of here? thoughts?

schemas/Clinical/index.js
```diff
@@ -278,6 +285,19 @@ const convertClinicalSubmissionErrorToGql = errorData => {
 };
 
 const resolvers = {
+  ClinicalRegistrationResp: {
```
joneubank: Real cool!

utils/restUtils.js
ciaranschutte: taking this specific case out as the default will return the response.
in the future we can add further logic for a 422 case if needed

ciaranschutte: @joneubank

hlminh2000:
looks like @jaserud is working on the same file with this PR: https://github.com/icgc-argo/platform-api/pull/51/files
Is it the same work?

ciaranschutte:
> looks like @jaserud is working on the same file with this PR: https://github.com/icgc-argo/platform-api/pull/51/files
> Is it the same work?

at a glance his work appears to be clinical submission vs this PR which is registration

jaserud:
^@hlminh2000 should be ok, this PR is modifying registration stuff. Mine is imodifying clinicalSubmission stuff only.

schemas/Clinical/index.js
```diff
@@ -44,6 +45,13 @@ const typeDefs = gql`
     alreadyRegistered: ClinicalRegistrationStats!
   }
 
+  type ClinicalRegistrationInvalid {
+    error: String
+    code: String
+  }
```
hlminh2000: to make UI life easier, probably a `programShortName: String` can help :) same with `ClinicalRegistrationData`

schemas/Clinical/index.js
```diff
@@ -312,16 +332,27 @@ const resolvers = {
...
+        // Success data is inside the key "registration", error data is in the root level
+        const data = { ...response.registration, errors: response.errors };
+        return convertRegistrationDataToGql(data);
+      } catch (e) {
+        // errors that don't go into error table
+        if (e.code) {
+          return { error: e.msg, code: ERROR_CODES[e.code] };
+        } else {
+          // catch all error
+          console.error('other error', e);
```
hlminh2000: there's a `winston` logger that can be used, import from `utils/logger.js` :) it'll comes with timestamp and log level too

schemas/Clinical/index.js
ciaranschutte: added `shortName` to both clinical responses

schemas/Clinical/index.js
ciaranschutte: use winston logger

schemas/Clinical/index.js
hlminh2000: I feel `shortName` is a little vague, what do you think of `programShortName`? :)

schemas/Clinical/index.js
hlminh2000: thank you!

schemas/Clinical/index.js
ciaranschutte: @hlminh2000 that works. to be clear, when do we use `shortName` and when do we use `programShortName`.

ending up pulling shortname from args
``` javascript
const {shortname} = args;
```
then renaming

hlminh2000: just cause the object type is `ClinicalRegistrationInvalid`, `shortName` sounds like `clinicalRegistrationInvalidShortName`, which isn't a thing :P just my feeling honestly

ciaranschutte: I see your thinking

PR #68, opened 2019-10-02
added createdAt

jaserud:
**Description of changes**

a lot of changes :P

No this is just adding a placeholder

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -293,7 +293,7 @@ const convertClinicalSubmissionEntityToGql = (type, entity) => {
       get(entity, 'dataErrors', []).map(error => convertClinicalSubmissionErrorToGql(error)),
     dataUpdates: () =>
       get(entity, 'dataUpdates', []).map(update => convertClinicalSubmissionUpdateToGql(update)),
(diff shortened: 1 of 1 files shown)


schemas/Clinical/index.js
```diff
@@ -292,6 +293,7 @@ const convertClinicalSubmissionEntityToGql = (type, entity) => {
       get(entity, 'dataErrors', []).map(error => convertClinicalSubmissionErrorToGql(error)),
     dataUpdates: () =>
       get(entity, 'dataUpdates', []).map(update => convertClinicalSubmissionUpdateToGql(update)),
+    createdAt: 'DateGoesHERE', // this is a place holder for now
```
hlminh2000: I think you want `new Date()` cause the `DateTime` type serializes a `Date` to a standard format

jaserud: hmm i tested that and it still worked :thinking:, any ways changed.

hlminh2000: hmmm gql might be trying to be graceful...

PR #70, opened 2019-10-03
add query for clinical type & createdAt

jaserud:
**Description of changes**

- added a query to ask clinical for clinical types list
- also take actual createdAt from clinical entity
- also also allow null values in clinical records
- changed shortName to programShortName for clinical submissions

Note: merge after https://github.com/icgc-argo/argo-clinical/pull/173

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -301,7 +301,7 @@ const convertClinicalSubmissionEntityToGql = (type, entity) => {
       get(entity, 'dataErrors', []).map(error => convertClinicalSubmissionErrorToGql(error)),
     dataUpdates: () =>
       get(entity, 'dataUpdates', []).map(update => convertClinicalSubmissionUpdateToGql(update)),
-    createdAt: entity.createdAt || null,
+    createdAt: entity.createdAt ? new Date(entity.createdAt) : null,
   };
 };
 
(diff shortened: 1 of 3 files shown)


schemas/Clinical/index.js
```diff
@@ -293,7 +301,7 @@ const convertClinicalSubmissionEntityToGql = (type, entity) => {
       get(entity, 'dataErrors', []).map(error => convertClinicalSubmissionErrorToGql(error)),
     dataUpdates: () =>
       get(entity, 'dataUpdates', []).map(update => convertClinicalSubmissionUpdateToGql(update)),
-    createdAt: new Date(), // this is a place holder for now
+    createdAt: entity.createdAt || null,
```
hlminh2000: is `entity.createdAt` a `Date` instance?

jaserud: yes, it is a Date instance in [clinical](https://github.com/icgc-argo/argo-clinical/blob/86c8626ed4e95804eba485ed74db33b2238db13f/src/submission/submission-entities.ts#L156)

hlminh2000: I think we'd be losing that context through the JSON serialization though :/

hlminh2000: maybe `entity.createdAt ? new Date(entity.createdAt) : null` to be safe? :P

PR #74, opened 2019-10-08
includes schemaErrors as a field in clinical entity

hlminh2000:
includes schemaErrors as a field in clinical entity

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -108,6 +108,7 @@ const typeDefs = gql`
     creator: String
     records: [ClinicalSubmissionRecord]!
     dataErrors: [ClinicalSubmissionError]!
+    schemaErrors: [ClinicalSubmissionError]!
     dataUpdates: [ClinicalSubmissionUpdate]!
     createdAt: DateTime
   }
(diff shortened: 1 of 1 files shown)


schemas/Clinical/index.js
```diff
@@ -257,20 +258,15 @@ const convertClinicalSubmissionDataToGql = data => {
...
-  }
-  // collect schema errors for each entity in dataErrors (not sure if this is OK??)
-  for (let clinicalType in schemaErrors) {
-    clinicalEntities.push(
-      convertClinicalSubmissionEntityToGql(clinicalType, {
-        dataErrors: schemaErrors[clinicalType],
-      }),
-    );
-  }
+  const clinicalEntities = Object.entries(submission.clinicalEntities).map(
```
jaserud: Hmm `Object.entries(submission.clinicalEntities)` can throw error. If submission doesn't exist in DB then default value is used from `const submission = get(data, 'submission', {});` where `submission.clinicalEntities` doesn't exist

hlminh2000: 👍 nice catch, thanks!

hlminh2000: @jaserud address! :)

jaserud: :+1:

PR #77, opened 2019-10-09
✨ associates clinical submission with program

hlminh2000:


schemas/Clinical/index.js
```diff
@@ -277,6 +278,7 @@ const convertClinicalSubmissionDataToGql = data => {
 
   return {
     id: submission._id || null,
+    programShortName,
```
jaserud: The submission returns programId which is programShortName, so we could use `submission.programId`. Unless there is some other reason to use the one passed in the query/mutation?? Hmm, I guess it is possible for submission.programId to be undefined if the submission doesn't exist.

hlminh2000: ah I see. I guess we can do `submission.programId || programShortName`? we do indeed want this to not be null so that client can associate mutation results with the data to update

hlminh2000: This also alows us to later on, stitch the relationship between `Program` and `ClinicalSubmissionData`, we'll need it in a soon to come feature

jaserud: :+1: nah that's fine then I just wanted to let you know that submission has it.

hlminh2000:
thanks @jaserud!

PR #81, opened 2019-10-10
includes schema errors in ClinicalSubmissionData

hlminh2000:
- includes schema errors in ClinicalSubmissionData
- creates `ClinicalEntityError` gql interface that is implemented by `ClinicalSubmissionError`
and `ClinicalSubmissionSchemaError`. `ClinicalSubmissionSchemaError` needs to have a `clinicalType` attached

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -276,21 +276,20 @@ const convertClinicalSubmissionDataToGql = (programShortName, data) => {
   const schemaErrors = get(data, 'schemaErrors', {});
   const fileErrors = get(data, 'fileErrors', []);
   // convert clinical entities for gql
-  const clinicalEntities = Object.entries(submission.clinicalEntities || {}).map(
-    ([clinicalType, clinicalEntity]) =>
-      convertClinicalSubmissionEntityToGql(
-        clinicalType,
-        clinicalEntity,
-        schemaErrors[clinicalType],
-      ),
-  );
 
   return {
     id: submission._id || null,
     programShortName,
     state: submission.state || null,
     version: submission.version || null,
-    clinicalEntities: clinicalEntities,
+    clinicalEntities: () =>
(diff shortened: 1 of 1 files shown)


schemas/Clinical/index.js
```diff
@@ -141,6 +143,25 @@ const typeDefs = gql`
...
+  type ClinicalSubmissionError implements ClinicalEntityError @cost(complexity: 5) {
+    type: String!
+    message: String!
+    row: Int!
+    field: String!
+    value: String!
+    donorId: String!
+  }
+
+  type ClinicalSubmissionSchemaError implements ClinicalEntityError {
```
joneubank: If youre going to move the complexity from the interface you need to specify it here.

schemas/Clinical/index.js
```diff
@@ -141,6 +143,25 @@ const typeDefs = gql`
     donorId: String!
   }
 
+  type ClinicalSubmissionError implements ClinicalEntityError @cost(complexity: 5) {
```
joneubank: rename to `ClinicalSubmissionDataError`?

schemas/Clinical/index.js
```diff
@@ -141,6 +143,25 @@ const typeDefs = gql`
...
+    type: String!
+    message: String!
+    row: Int!
+    field: String!
+    value: String!
+    donorId: String!
+  }
+
+  type ClinicalSubmissionSchemaError implements ClinicalEntityError @cost(complexity: 10) {
```
hlminh2000: added query cost to this, making it 10 since there's a bit more wrangling to get these compared to `ClinicalSubmissionError`

schemas/Clinical/index.js
hlminh2000: just realized this right after asking for review. I've made it 10 cause there's some wrangling to flatten the upstream data :)

schemas/Clinical/index.js
hlminh2000: addressed :)

PR #82, opened 2019-10-10
Feature/57 all template zip

jaserud:
**Description of changes**

- added support for all.zip (zip file with all schema templates) pass through from clinical
- added new endpoint in swagger
- catch errors

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature


jaserud:
sankyu

PR #83, opened 2019-10-10
✨ includes empty clinical entities in the ClinicalSubmissionData type

hlminh2000:
You know what's up :P

We decided to move forward with this, using the solution from gateway


joneubank:
maybe you wanna increase the complexity associated with this return type now?

hlminh2000:
> maybe you wanna increase the complexity associated with this return type now?

added cost of `20` to the `clinicalEntities` field :)

PR #85, opened 2019-10-11
🐛 ClinicalSubmissionSchemaError value field can be null

hlminh2000:
turns out for schema errors, `value` can be null


joneubank:
makes sense... is it possible for donor_id to be null as well if that wasnt provided in the file?

hlminh2000:
> makes sense... is it possible for donor_id to be null as well if that wasnt provided in the file?

Thanks, I just tried that, those become `""`... maybe it makes better sense to default value to `""`?

joneubank:
Whatever is less likely to cause an error in the UI. I think its clear that a "" value means nothing was provided in the .tsv

hlminh2000:
> Whatever is less likely to cause an error in the UI. I think its clear that a "" value means nothing was provided in the .tsv

ok I'll keep it as `String!` and enforce empty to `""` then, just to avoid an API change

PR #90, opened 2019-10-18
Add Reopen Clinical Submission Mutation

jaserud:
**Description of changes**

- added support for reopen
- also removed .zip from all template endpoint/path

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

### DIFF
--- a/routes/clinical.js
+++ b/routes/clinical.js
@@ -10,7 +10,9 @@ var router = express.Router();
 // so that the user is automatically prompted to save the file from their browser.
 const apiRoot = urlJoin(CLINICAL_SERVICE_ROOT, SUBMISSION_TEMPLATE_PATH);
 
-// This is mostly passthrough, so file names and response should be set in clincial
+// This is mostly passthrough, so file names and response will be set in clincial
+// 'all'.zip or 'all' will retrieve the zip file with all templates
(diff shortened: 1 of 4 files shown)


routes/clinical.js
```diff
@@ -10,7 +10,9 @@ var router = express.Router();
 // so that the user is automatically prompted to save the file from their browser.
 const apiRoot = urlJoin(CLINICAL_SERVICE_ROOT, SUBMISSION_TEMPLATE_PATH);
 
-// This is mostly passthrough, so file names and response should be set in clincial
+// This is mostly passthrough, so file names and response will be set in clincial
```
jaserud: @hlminh2000 not sure if this affects you right now...

hlminh2000: sounds backward compatible so shouldn't be a breaking change for me. Though does this explanation appear in swagger somewhere? :P

jaserud: no lol, guess I could put in something

hlminh2000: yeah I just figured as a consumer of the API I'd prefer to find this information there than digging into source code :P thanks!

jaserud: I removed the all.zip it was more confusing

PR #94, opened 2019-10-22
💥 make createProgram mutation return new program created

hlminh2000:
- needed for https://github.com/icgc-argo/platform-ui/pull/783
- updates `createProgram` mutation to return the program created

### DIFF
--- a/schemas/Program/index.js
+++ b/schemas/Program/index.js
@@ -160,7 +160,7 @@ const typeDefs = gql`
     For lists (Cancer Type, Primary Site, Institution, Regions, Countries) the entire new value must be provided, not just values being added.
     Returns Program object details of created program
     """
-    createProgram(program: ProgramInput!): [Program!] @cost(complexity: 20)
+    createProgram(program: ProgramInput!): Program! @cost(complexity: 40)
 
     """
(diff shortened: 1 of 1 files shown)


joneubank:
There is a limitation from ProgramService which makes this logic fail - specifically there is a timing issue where createProgram request returns success before completing writing the data, and calling listPrograms immediately after will result in the new program being missing from the list. We used to have it this way and resulted in issues where the gateway reported a failure to create despite it eventually working.

For now, I encourage us to leave the createProgram mutation as it is and handle updating the List in the UI via a separate query.

hlminh2000:
Thanks @joneubank I'll update the UI to work with that logic :)

For future I'd still like to have this driven from the gateway if possible, because the need to update UI state in other places may show up again, which makes it difficult to keep track of what query to run when.

joneubank:
Agreed, that is an ideal behaviour.. this particular case will require updates to the programService and is limited just to a rare DCC action... we should keep a ticket for these changes in case we get the time but I wouldn't prioritize it just yet. Future interfaces should definitely be built with this style of behaviour in mind (returning full objects/lists after mutation).

PR #101, opened 2019-10-31
Remove Clinical Error Messages

joneubank:
**Description of changes**

Removing error messages for clinical, the messages are constructed by the clinical service now and the gateway simply puts them into the gql response.


joneubank:
Do not deploy before https://github.com/icgc-argo/argo-clinical/pull/222

PR #115, opened 2019-11-22
fix registration query

jaserud:
**Description of changes**
 looks like registration query was returning a empty registration, fixed it :+1: 

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -415,7 +415,7 @@ const resolvers = {
       const { shortName } = args;
 
       const response = await clinicalService.getRegistrationData(shortName, Authorization);
-      return convertRegistrationDataToGql({ ...response, shortName });
+      return convertRegistrationDataToGql({ registration: response, shortName });
     },
     clinicalSubmissions: async (obj, args, context, info) => {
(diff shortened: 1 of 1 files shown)


schemas/Clinical/index.js
```diff
@@ -415,7 +415,7 @@ const resolvers = {
       const { shortName } = args;
 
       const response = await clinicalService.getRegistrationData(shortName, Authorization);
-      return convertRegistrationDataToGql({ ...response, shortName });
+      return convertRegistrationDataToGql({ registration: response, shortName });
```
hlminh2000: is the `shortName` supposed to go under registration as well?

jaserud: it is their inside registration from clinical, but it looks like shortName is used by `convertRegistrationDataToGql`, I tried to keep my changes to a minimum :slightly_smiling_face:

hlminh2000: in case of no registration, the UI still needs the `shortName` there because it's used as the cache key, so I think that's why we include the `shortName` from `args`

PR #118, opened 2019-11-28
updates Ego api to match with latest release

hlminh2000:
- updates ego according to these changes here: https://github.com/overture-stack/ego/pull/420
- adds some [JSDoc](https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#supported-jsdoc) to provide type info (I think we should do more of this in the gateway, it's not perfect but better than nothing if we're not getting the luxury of TypeScript)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-  "javascript.validate.enable": false,
-  "flow.useNPMPackagedFlow": true,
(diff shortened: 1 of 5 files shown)


hlminh2000:
if you set the following in `.vscode/settings.json`, vscode will also highlight errors:
```
{
  "javascript.implicitProjectConfig.checkJs": true
}
```

PR #123, opened 2019-12-04
Add 5xx error handling

jaserud:
**Description of changes**

- added generic handle 5xx errors
- throw ApoolloError (ServerError wasn't defined)
- some cleanup

this is to handle new error from clinical [(PR#266)](https://github.com/icgc-argo/argo-clinical/pull/266) and possibly else where

**Type of Change**

- [ ] Bug
- [x] Styling
- [x] New Feature

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -1,8 +1,6 @@
 import { gql, AuthenticationError } from 'apollo-server-express';
 import { makeExecutableSchema } from 'graphql-tools';
 import get from 'lodash/get';
-import omit from 'lodash/omit';
-import flattenDeep from 'lodash/flattenDeep';
 
 import createEgoUtils from '@icgc-argo/ego-token-utils/dist/lib/ego-token-utils';
 
@@ -12,8 +10,6 @@ const TokenUtils = createEgoUtils(EGO_PUBLIC_KEY);
 
 import costDirectiveTypeDef from '../costDirectiveTypeDef';
 import clinicalService from '../../services/clinical';
-import customScalars from '../customScalars';
-import { ERROR_CODES } from '../../constants/clinical';
 import logger from '../../utils/logger';
 
 const typeDefs = gql`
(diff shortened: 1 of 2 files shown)


schemas/Clinical/index.js
```diff
@@ -1,8 +1,6 @@
 import { gql, AuthenticationError } from 'apollo-server-express';
```
jaserud: all changes in this file is from prettier

utils/restUtils.js
```diff
@@ -1,8 +1,21 @@
...
+  // Generic handle 5xx errors
+  if (response.status >= 500 && response.status <= 599) {
+    let responseBody;
+    try {
+      responseBody = await response.json();
+    } catch {
+      responseBody = { message: '' };
+    }
+    logger.debug(`Server 5xx response: ${JSON.stringify(responseBody)}`);
+    throw new ApolloError(responseBody.message || '', response.status);
```
blabadi: I don't know if passing the downstream error message through to the caller is a good idea

jaserud: fair enough, how about the http error code

blabadi: depends what the UI is expecting honestly and what the graphql standard errors for such cases. I'm not sure if the response status code is going to be interpreted by them .. Jon and Minh can provide better opinion

jaserud: I've removed all error info from going to downstream... we can add later if needed.

utils/restUtils.js
```diff
@@ -1,8 +1,22 @@
...
+import logger from './logger';
 /*
 convert the REST status codes to GQL errors, or return the response if passing
 */
 export const restErrorResponseHandler = async response => {
+  // Generic handle 5xx errors
+  if (response.status >= 500 && response.status <= 599) {
+    let responseBody;
+    try {
+      responseBody = await response.json();
```
hlminh2000: since we're not expecting to parse `responseBody` for anything, I think `.text()` is safer. The only reason `.json()` would fail is because rest APIs are not guaranteed to return json, which is an assumption we can safely bypass with `.text()`.

utils/restUtils.js
```diff
@@ -1,8 +1,22 @@
...
+    } catch {
+      responseBody = response;
+    }
+    logger.debug(`Server 5xx response status: ${response.status}`);
+    logger.debug(`Server 5xx response body: ${JSON.stringify(responseBody)}`);
```
hlminh2000: with `.text()`, no need to stringify it back

PR #125, opened 2019-12-17
added get clinical schema version

jaserud:
**Description of changes**

- added new query for UI to get schema version
- also added invalid_by_migration to code list

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

### DIFF
--- a/services/clinical/index.js
+++ b/services/clinical/index.js
@@ -73,6 +73,17 @@ const getClinicalSubmissionTypesList = async () => {
   return response;
 };
 
+const getClinicalSubmissionSchemaVersion = async () => {
+  const url = `${CLINICAL_SERVICE_ROOT}/submission/schema`;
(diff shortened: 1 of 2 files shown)


services/clinical/index.js
```diff
@@ -73,6 +73,17 @@ const getClinicalSubmissionTypesList = async () => {
   return response;
 };
 
+const getClinicalSubmissionSchemaVersion = async () => {
```
hlminh2000: let's add some jsdoc to new utilities :) https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#supported-jsdoc

vscode has some autocomplete built-in so that should help ;)

jaserud: in the end I added for this one only because I don't remember what the return type is for the other ones :stuck_out_tongue:

PR #132, opened 2020-01-28
fix null value not being checked correctly

jaserud:
**Description of changes**

`null` was being returned by GQL for `value`, but it was reporting other fields as being null.

**Type of Change**

- [X] Bug
- [ ] Styling
- [ ] New Feature

### DIFF
--- a/schemas/Clinical/index.js
+++ b/schemas/Clinical/index.js
@@ -387,7 +387,7 @@ const convertClinicalSubmissionDataErrorToGql = errorData => {
     donorId: get(errorData, 'info.donorSubmitterId', ''),
 
     // errorData.info.value may come back as null if not provided in uploaded file
-    value: get(errorData, 'info.value', ''),
+    value: get(errorData, 'info.value', '') || '',
   };
 };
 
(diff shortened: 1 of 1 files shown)


schemas/Clinical/index.js
```diff
@@ -387,7 +387,7 @@ const convertClinicalSubmissionDataErrorToGql = errorData => {
     donorId: get(errorData, 'info.donorSubmitterId', ''),
 
     // errorData.info.value may come back as null if not provided in uploaded file
-    value: get(errorData, 'info.value', ''),
+    value: get(errorData, 'info.value', '') || '',
```
hlminh2000: do we expect the same for `donorId`?

jaserud: that shouldn't be possible because every record has a submitterDonorId, clinical wouldn't have accepted the file if it was missing. But I guess it doesn't hurt to add that check.

hlminh2000: the gql schema defines the field as a `String!`, so regardless of what the possibility, we have to ensure a string comes through to it. If such guarantee is not the case, it should be typed as `String` :)

jaserud: agreed

PR #146, opened 2020-02-20
🐛 updates to match new ego api

hlminh2000:


joneubank:
This will break for users with >1000 keys. Shouldn't we call all pages to find existing tokens?

services/ego/index.js
```diff
@@ -113,27 +114,27 @@ const getScopes = async (userName, Authorization) => {
 };
 
 /**
- * @param {Array<EgoAccessKeyResponse>} keys
+ * @param { EgoAccessKeyResponse[] } keys
```
joneubank: Why wouldn't we pass a list of strings of the key names instead of the response? As written here, we couldn't reuse this service to delete (revoke) specific keys through a mutation...

hlminh2000: that's how it was before so I just kept it to avoid making mistakes and introducing bugs :P

PR #171, opened 2020-03-20
Bug/donor submission schema revision

hlminh2000:
Added a bunch of docs and fixed some typos

@rosibaj I've added a bunch of docs to the schema, we can get this out to QA and if there's anything that doesn't make sense, please feel free to make some PR to update :) Just wanted to tag you to this for reference. 

Also wanted to get your thoughts on the name for the enums `DonorMolecularDataReleaseStatus` and `DonorMolecularDataProcessingStatus`, are they accurate?

### DIFF
--- a/src/schemas/ProgramDonorSummary/resolvers.ts
+++ b/src/schemas/ProgramDonorSummary/resolvers.ts
@@ -19,13 +19,6 @@ const programDonorSummaryEntriesResolver: GraphQLFieldResolver<
 
   return [];
 };
-
-const totalDonorsCountResolver: GraphQLFieldResolver<
-  ProgramDonorSummaryStats,
-  GlobalGqlContext
-> = () => {
-  return 0;
-};
 const fullyReleasedDonorsCountResolver: GraphQLFieldResolver<
   ProgramDonorSummaryStats,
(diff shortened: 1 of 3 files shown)


src/schemas/ProgramDonorSummary/resolvers.ts
```diff
@@ -7,25 +7,18 @@ const programDonorSummaryEntriesResolver: GraphQLFieldResolver<
...
 > = (source, args, context): DonorSummaryEntry[] => {
-  const { programId } = args;
+  const { programShortName } = args;
 
   console.log('args: ', args.filters);
 
   return [];
 };
-
-const totalDonorsCountResolver: GraphQLFieldResolver<
```
jaserud: couldn't we delete the other donorCount resolvers too?

rosibaj:
@hlminh2000  I believe that these are accurate on naming

src/schemas/ProgramDonorSummary/resolvers.ts
hlminh2000: yes we can! sorry forgot to, and already did that in another coming PR, so will skip for this PR :) thanks!

PR #186, opened 2020-04-09
check for no alias and feature flag empty resolvers

jaserud:
**Description of changes**
- added empty resolvers behind a feature flag
- added check for alias before searching

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -59,7 +59,7 @@ const init = async () => {
     userSchema,
     programSchema,
     clinicalSchema,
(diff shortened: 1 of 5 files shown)


src/app.ts
```diff
@@ -59,7 +59,7 @@ const init = async () => {
     userSchema,
     programSchema,
     clinicalSchema,
-    ...(PROGRAM_DASHBOARD_SUMMARY_ENABLED ? [await ProgramDashboardSummarySchema()] : []),
```
jaserud, line 62: I'm pretty sure UI would break if this stayed true.

PR #192, opened 2020-04-24
Feat/secure dashboard

hlminh2000:
**Description of changes**

- installs latest `@icgc-argo/ego-token-utils`
- converts clinical and ego resolvers to typescript and fixes compatibility issue with latest util
- implements access control for the dashboard data

**Type of Change**

- [x] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


src/schemas/Clinical/index.ts
```diff
@@ -459,20 +553,33 @@ const resolvers = {
...
         registrationId,
         Authorization,
       );
+      //@ts-ignore consult Jaser on modifying clearRegistrationData or remove this check altogether
       if (response.error) {
+        //@ts-ignore
         throw new UserInputError(response.message);
       }
       return true;
     },
```
hlminh2000, line 577: @jaserud I don't think this block is ever called because the response isn't returned by `clinicalService.clearRegistrationData`. Is this correct?

src/schemas/ProgramDonorSummary/resolvers/index.ts
```diff
@@ -3,13 +3,76 @@ import { GlobalGqlContext } from 'app';
...
+        permissions,
+        programId: args.programShortName,
+      }) ||
+      canReadProgramData({
+        permissions,
+        programId: args.programShortName,
+      });
+
+    const authorized =
+      egoToken && isValidJwt(egoToken) && !isExpiredToken(decodedToken) && hasPermission;
```
jaserud, line 52: could use the const `isExpired` no?

jaserud, line 52: could use the const and check `!isExpired`...??

src/schemas/Clinical/index.ts
```diff
@@ -0,0 +1,454 @@
...
+    ) => {
+      const { Authorization } = context;
+      const { shortName, registrationId } = args;
+      const response = await clinicalService.clearRegistrationData(
+        shortName,
+        registrationId,
+        Authorization,
+      );
+      //@ts-ignore consult Jaser on modifying clearRegistrationData or remove this check altogether
+      if (response.error) {
```
jaserud, line 329: Took a look at the clinical code, there are some cases where response can be an error.... If the versionId doesn't match, submission doesn't exist, or if the submission is pending approval

jaserud: oops that submisison

jaserud: clear registration doesn't return anything if success, just code 200. I think the restErrorResponseHandler will take care of any other case where there is an error, but that won't make it to UI so yea I guess you could remove this.

hlminh2000: 👍 cool thanks!

src/schemas/ProgramDonorSummary/resolvers/index.ts
hlminh2000: yes! :P me derped

PR #203, opened 2020-05-06
✨ Init sample es index script

hlminh2000:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] Styling
- [ ] New Feature

- [ ] Matches design:

  - font size, weight, colour


hlminh2000:
@anncatton we can add that case by case for each index. The existing script is really only for the dashboard data, and only for dev, so doesn't need too much engineering IMP

PR #209, opened 2020-05-14
Mutation for making a Jira Service Desk Request

wajiha-oicr:
**Description of changes**

- Wrote Mutation that will accept information from the contact us page and send it to Jira's REST API endpoint for creating a service desk request

- Wrote some other (unused) requests that go through a different process to obtain a clientID so the request can be filed the way the [api docs stated 
](https://developer.atlassian.com/cloud/jira/service-desk/rest/#api-rest-servicedeskapi-request-post ) ( they said the 'raiseOnBehalfOf' field needs to be given a client id, although from testing, an email works just fine)

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

- [x] Matches design:

  - font size, weight, colour (yup the colours all perfect)


src/schemas/Helpdesk/gqlTypeDefs.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export default gql`
+  type TicketCreationResponse {
+    requestOk: Boolean!
+    statusText: String!
+  }
+
+  type Mutation {
+    createJiraTicket(
+      dropdownValue: String!
```
hlminh2000, line 11: `dropdownValue` doesn't really mean much when looking at the schema. Knowing what it is though, something like `messageCategory` or even `summaryPrefix` seems to make more sense. 

Another note on this is, since it's a limited list of things, this might make sense as an enum.

src/schemas/Helpdesk/jiraRequests.ts
```diff
@@ -0,0 +1,208 @@
...
+
+        try {
+          const isAlreadyCustomer = errorData.i18nErrorMessage.parameters.includes(
+            EMAIL_ALREADY_EXISTS_MESSAGE,
+          );
+          if (isAlreadyCustomer) {
+            return await getCustomerId(email);
+          }
+          else {
+            console.debug('An Error Occured, and the error body couldnt be correctly parsed.');
```
hlminh2000, line 147: let's use `logger` for consistency :)

src/schemas/Helpdesk/jiraRequests.ts
```diff
@@ -0,0 +1,208 @@
...
+      headers: await getJiraRequestHeader(),
+      body: bodyData,
+    })
+
+    .then(restErrorResponseHandler)
+  
+    .then( async response => {
+      return {requestOk: response.ok, statusText: response.statusText}
+
+    });
```
hlminh2000, line 84: doesn't this endpoint return a json body? if so, it's more meaningful to return `response.json()`, because `response.ok` and `response.statusText` are pretty much just http implementation details. The graphql type can just expose a subset of the data returned if it's a lot of uninteresting data, but something meaningful is still better :)

src/schemas/Helpdesk/jiraRequests.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const bodyData = JSON.stringify({
+    displayName: displayName,
+    email: email,
+  });
+
+
+  const response = await fetch(url, {
+    method: 'post',
+    headers: await getJiraRequestHeader(),
```
hlminh2000, line 128: reaching out to vault on every request seems unnecessary, we can just get it once when the app starts up and keep it in memory so there's less reliance on vault during he runtime. 

If we wrap all the functions in this file into a `JiraClient` type object, that is instantiated using a `createJiraClient` function, then `createJiraclient` can get the credential once and this method can use that credential without reaching vault everytime :)

src/schemas/Helpdesk/jiraRequests.ts
wajiha-oicr: I tried that initially and noticed that for failed attempts ( i.e. unauthenticated) the returned json body would be considered invalid, and there would be a parsing error thrown

hlminh2000: this is the authentication of our app against JIRA right? If so this error represents a bug in the system rather than a meaningful error that user should be expected to handle.

The best thing to do there would be to test the credential at the app's initialization time and fail the app if the credential isn't correct. In the elasticsearch client case, we do this by pinging elasticsearch using the instantiated client, if the ping fails then we know there's a bug in the configuration and the app should not be able to start up.

src/schemas/Helpdesk/gqlTypeDefs.ts
wajiha-oicr: I was thinking of doing the enum, but thought it would have to be defined in the UI end of things anyway, and so wouldn't adding it here as well required twice the maintenance, should something change?

hlminh2000: Since this mutation is publicly available, the UI is not guaranteed to be the only place requests are coming from, it's a good vandalism mitigation measure :) also prevent us making mistakes if we need to call this from other places too

wajiha-oicr:
@hlminh2000 thanks for the great feedback! I've made and tested ( use_vault = false though) the changes here https://github.com/icgc-argo/platform-api/pull/209/commits/f85d8f24c980a0d8f1f5872386c1299315f5edd2

src/schemas/Helpdesk/index.ts
```diff
@@ -0,0 +1,73 @@
...
+        emailAddress: string;
+        requestText: string;
+        displayName: string;
+      },
+      context: GlobalGqlContext,
+    ) => {
+      const { messageCategory, emailAddress, requestText, displayName } = args;
+
+      const messageCategoryKey = messageCategory as keyof typeof Category;
+      const jiraClient = await createJiraClient();
```
hlminh2000, line 56: This looks like it's creating a new jira client on every request, which would still result in a request to vault everytime. To fix this, we can update the export of this file to be something like this:
```
async () => {
  const jiraClient = await createJiraClient();
  makeExecutableSchema({
    typeDefs,
    resolvers: createResolvers(jiraClient)
  })
}
```
this will allow us to create one jira client that is then used for the entire runtime of the app :)

src/schemas/Helpdesk/gqlTypeDefs.ts
```diff
@@ -0,0 +1,38 @@
...
+      emailAddress: String!
+      requestText: String!
+      displayName: String
+    ): TicketCreationResponse!
+  }
+
+  # one query declaration required
+  type Query {
+    _dummy: String
+  }
```
hlminh2000, line 37: I think this whole chunk can be removed eh? :P

src/schemas/Helpdesk/gqlTypeDefs.ts
```diff
@@ -0,0 +1,38 @@
+import { gql } from 'apollo-server-express';
+
+export default gql`
+  enum Category {
```
hlminh2000, line 4: I would name this someting like `JiraTicketCategory`, only because when this gets merged into the larger schema, `Ticket` is a little generic and creates room for namespace conflict. 
It's an unfortunate thing that we can't really do anonymous types or namespace them in GraphQl so we have this tricky split situation eh

src/schemas/Helpdesk/gqlTypeDefs.ts
wajiha-oicr: it breaks without at least one query :/ 

`Error: Query root type must be provided.`

PR #293, opened 2020-08-17
add datawarnings to clinical entity data

blabadi:
add datawarnings to clinical entity data
use stub recaptcha client if flag disabled

**Description of changes**

**Type of Change**

- [x] New Feature

### DIFF
--- a/src/schemas/Helpdesk/index.ts
+++ b/src/schemas/Helpdesk/index.ts
@@ -21,7 +21,7 @@ import typeDefs from './gqlTypeDefs';
 import { GlobalGqlContext } from 'app';
 import { makeExecutableSchema } from 'graphql-tools';
 import { createJiraClient, JiraClient } from './jiraRequests';
-import { FEATURE_HELP_DESK_ENABLED } from 'config';
+import { FEATURE_HELP_DESK_ENABLED, DEV_RECAPTCHA_DISABLED } from 'config';
 import { ApolloError, UserInputError } from 'apollo-server-express';
 import createReCaptchaClient, { ReCaptchaClient, createStubReCaptchaClient } from 'services/reCaptcha';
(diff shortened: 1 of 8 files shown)


src/schemas/Helpdesk/index.ts
```diff
@@ -128,16 +124,11 @@ const createDisabledResolvers = (reCaptchaClient: ReCaptchaClient) => ({
...
-      throw new ApolloError(
-        `FEATURE_HELP_DESK_ENABLED is ${FEATURE_HELP_DESK_ENABLED} in this instance of the gateway`,
-      );
-    },
   },
 });
 
 export default async () => {
-  const reCaptchaClient = await createReCaptchaClient(); // this client is created regardless of FEATURE_HELP_DESK_ENABLED so its vault integration can be tested in dev environments
+  const reCaptchaClient = FEATURE_HELP_DESK_ENABLED ? await createReCaptchaClient(): await createStubReCaptchaClient(); // this client is created regardless of FEATURE_HELP_DESK_ENABLED so its vault integration can be tested in dev environments
```
hlminh2000, line 131: If we use a stub client then it's not really testable in dev anyway so the comment would be irrelevant. 

Basically the situation is, outside of prod, we don't have a JIRA to integrate with, so we left `FEATURE_HELP_DESK_ENABLED` out in other environments. But there, we still wanted to be able to test out reCaptcha, which is why reCaptcha integration was left in place in the `createDisabledResolvers`.

I think it'd be better to introduce a `DEV_RECAPCHA_DISABLED` flag for local dev and use it here so we retain the ability to test reCaptcha. No need to update any existing infra stuff because it'd be `false` by default. We can also update the error message thrown by `createReCaptchaClient` to mention this new flag... What do you think?

blabadi: yeah that makes sense

PR #326, opened 2020-10-21
:bug: local dev elasticsearch secret support

hlminh2000:
- allows specifying ES credential through env


ciaranschutte (approved):
Can you add the elastic search user name and pass to schema.env please :)

PR #329, opened 2020-10-26
Feat/entities

hlminh2000:
- minor refactoring to get a function set up for reuse
- implements single entity retrieval endpoint

### DIFF
--- a/src/routes/file-storage-api/utils.ts
+++ b/src/routes/file-storage-api/utils.ts
@@ -1,0 +1,16 @@
+import { Client } from "@elastic/elasticsearch";
+import { ARRANGER_FILE_CENTRIC_INDEX } from "config";
+import esb from "elastic-builder";
+import { EsFileCentricDocument } from "utils/commonTypes/EsFileCentricDocument";
+
+export const getEsFileDocumentByObjectId = (esClient: Client) => (objectId: string) => {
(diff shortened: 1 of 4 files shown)


src/routes/file-storage-api/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+export const getEsFileDocumentByObjectId = (esClient: Client) => (objectId: string) => {
+  const objectIdField: keyof EsFileCentricDocument = 'object_id';
+  return esClient
+    .search({
+      index: ARRANGER_FILE_CENTRIC_INDEX,
+      body: esb
+        .requestBodySearch()
+        .query(esb.boolQuery().must(esb.termQuery(objectIdField, objectId))),
+    })
+    .then(res => res.body.hits.hits[0]?._source as EsFileCentricDocument | undefined);
```
ciaranschutte, line 15: @hlminh2000 just making sure `.hits.hits` isn't a typo as a repeated property

ciaranschutte (approved):

src/routes/file-storage-api/utils.ts
hlminh2000: nope, it actually isn't :P ES is just verbose like that...

PR #339, opened 2020-11-10
🐛 make hasSufficientDacoAccess and hasSufficientProgramMembership separate

hlminh2000:


ciaranschutte (approved):

src/routes/file-storage-api/accessValidations.ts
```diff
@@ -72,24 +75,23 @@ export const hasSufficientProgramMembershipAccess = (config: {
...
+export const hasSufficientDacoAccess = (config: {
+  scopes: PermissionScopeObj[];
+  file: EsFileCentricDocument;
+}): boolean => {
   const dacoScopes = config.scopes.filter(({ policy }) => policy === EGO_DACO_POLICY_NAME);
-  return dacoScopes.length > 0 && dacoScopes.every(scope => scope.permission === PERMISSIONS.READ);
+  const userHasDacoAccess =
+    dacoScopes.length > 0 && dacoScopes.every(scope => scope.permission === PERMISSIONS.READ);
+  return (
+    config.file.file_access === FILE_ACCESS.PUBLIC ||
```
ciaranschutte, line 92: @hlminh2000 do we need to check for `file` property on config? 
```
config.file && config.file.file_access  ...
```

hlminh2000: `file` is a required field on `config`, so you can't call `hasSufficientDacoAccess` without passing in a `file`. So no need for that. In the UI we would have to, because we don't have strict null check there

PR #359, opened 2021-01-18
✨ 282 - add banners to API

samrichca:
add banner messages to API


src/schemas/Banner/index.ts
```diff
@@ -0,0 +1,29 @@
+import { gql } from 'apollo-server-express';
+import { makeExecutableSchema } from 'graphql-tools';
+
+const bannersStr = process.env.BANNERS || '';
+const bannerArray = JSON.parse(bannersStr);
```
joneubank, line 5: If the JSON is poorly formed, will this prevent the server from starting?

I just want to confirm that the server will still spin up (and just not show poorly written banners) if we typo the banner data.

samrichca: that's a very good point

I added error handling for 
- no banners
- there's valid json, but it's not an array
- poorly formed json

src/schemas/Banner/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  try {
+    const bannersStr = process.env.BANNERS || '';
+    const banners = JSON.parse(bannersStr);
+    if (!Array.isArray(banners)) {
+      throw new Error('Banners need to be an array');
+    } 
+    return banners;
+  } catch (e) {
+    console.log('Banners error: ', e.name, e.message)
+    return [];
```
joneubank, line 14: Thanks for handling this. Returning empty array is perfect.

joneubank (approved):

src/schemas/Banner/index.ts
```diff
@@ -0,0 +1,91 @@
...
+        const expectedType = bannerFieldsAllowed[key];
+        const checkType = expectedType === typeof value;
+        if (!checkType) {
+          throw new Error(`${key} should be a ${expectedType} in ${errorTitle}`);
+        }
+      }
+      
+      const checkLevel = bannerLevelsAllowed.includes(banner.level);
+      if (!checkLevel) {
+        throw new Error(`Invalid level "${banner.level}" in ${errorTitle}`);
```
joneubank, line 56: If one banner is bad but the others are good, maybe we just want to exclude the broken banner and show the others?

src/schemas/Banner/index.ts
```diff
@@ -0,0 +1,91 @@
...
+const bannerLevelsAllowed = [
+  'error',
+  'info',
+  'warning',
+];
+
+const getBanners = () => {
+  try {
+    const bannersStr = process.env.BANNERS || '';
+    const bannersParsed = JSON.parse(bannersStr);
```
ciaranschutte, line 30: Can we just use Typescript and type our `bannerKeys` here and have it cover a lot of the error handling?
```
const bannersParsed: Array<Banners>  = JSON.parse(bannersStr)
```

src/schemas/Banner/index.ts
ciaranschutte: @joneubank thoughts on using TS and just making sure the banner is correct at buildtime?

ciaranschutte:
scratch my TS comment, pulling the banner in at runtime, no type checking for malformed data at that point

PR #363, opened 2021-01-27
Rc/3.13.1

samrichca:
## New Features

* _banners_ type added to gql schema, data provided through env config

### DIFF
--- a/src/schemas/Banner/index.ts
+++ b/src/schemas/Banner/index.ts
@@ -1,0 +1,100 @@
+import { gql } from 'apollo-server-express';
+import { makeExecutableSchema } from 'graphql-tools';
+import bannerTI from "./banner-ti";
+import { createCheckers } from "ts-interface-checker";
+import logger from '../../utils/logger';
+
+const { BannerTI } = createCheckers(bannerTI);
+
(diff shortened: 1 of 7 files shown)


joneubank (approved):

src/schemas/Banner/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  title: string;
+}
+
+const allowedFields = [
+  'dismissable',
+  'id',
+  'level',
+  'message',
+  'title',
+];
```
hlminh2000, line 31: if you type this as `AllowedFields[]`, it'll help avoid mistakes in forgetting to add fields when this changes ;)

hlminh2000: actually, `AllowedFields` can just be `keyof Banner` too

ciaranschutte: god definition i like that

samrichca: i updated `AllowedFields` to be `keyof Banner` and added a comment about the other issue

tl;dr we're doing runtime checks because there's no other data validation for the banners and we don't want the build to crash

PR #382, opened 2021-03-28
Donor Aggregator Route with DCC Admin restrictions

joneubank:
**Purpose of this PR**

A REST endpoint is created to allow DCC Admin users to trigger a re-index of the donor summary index by the donor submission aggregator.

The new endpoint is at `/donor-aggregator/sync`, and the relevant method details have been added to the swagger documentation.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

**Summary of Changes**
Adding a bit more detail here because I needed to do some refactoring in order to handle the authorization requirements for the new route:
- Added router for donor-aggregator: `/donor-aggregator`
  - includes one endpoint: `POST /sync`
- Add swagger documentation for the new endpoint as a partial document
- Add a donor-aggregator service to make requests to the donor-submission-aggregator.
- Updates the code to generate the swagger docs page `src/routes/api-docs.ts` to take any number of partial swagger files and merge them into the final document
- Removed `isExpiredToken` from the `egoTokenUtils` object
  - This method had an implementation that was not correct, same issue as here: https://github.com/icgc-argo/donor-submission-aggregator/pull/173
  - All uses of this method were replaced with `isValidJwt` which includes a check for expiry.
- Refactored the `authenticatedRequestMiddleware` and the authorization validation methods to move these to common files.
  - These are used to enforce that all requests to the new endpoint are from a DCC admin.

One last note: the auth is actually handled twice, once in the express router by the authentication middleware and then again by the service that handles the requests to the aggregation service. This is redundant, but better safe than sorry. I did this to ensure that the service itself has authorization checks in case anything else calls it in the future, but I also think it is good practice to have the endpoint specify the rules for who can access.


src/schemas/ProgramDonorSummary/gqlTypeDefs.ts
```diff
@@ -233,6 +233,11 @@ export default gql`
     Number of donors registered to the program who currently has no released genomic file
     """
     noReleaseDonorsCount: Int!
+
+    """
+    Date of the most recent update to the donor summary index for this program. Can be null if no documents for this program
+    """
+    lastUpdate: DateTime
```
joneubank, line 240: This was a low cost add to the donor summary stats, getting the most recent donor updated date for a program.

src/schemas/ProgramDonorSummary/resolvers/index.ts
```diff
@@ -89,7 +70,7 @@ const resolveWithProgramAuth = <ResolverType = GraphQLFieldResolver<unknown, unk
 
 const createResolvers = async (
   esClient: Client,
-): Promise<IResolvers<unknown, GlobalGqlContext>> => {
+): Promise<IResolvers<ProgramDonorSummaryStatsGqlResponse, GlobalGqlContext>> => {
```
joneubank, line 73: I poked around at the types involved in resolvers and this will be relevant if we add a resolver for a specific field instead of the entire type. This isn't needed right now but I don't want to lose this detail for if this does get modified later.

src/schemas/ProgramDonorSummary/resolvers/summaryEntries.ts
```diff
@@ -46,7 +44,15 @@ const programDonorSummaryEntriesResolver: (
     sorts: DonorSummaryEntrySort[];
     filters: ProgramDonorSummaryFilter[];
   }
-> = esClient => async (source, args, context): Promise<DonorSummaryEntry[]> => {
+>;
+
+const programDonorSummaryEntriesResolver: (
+  esClient: Client,
+) => DonorEntriesResolverType = esClient => async (
```
joneubank, line 51: Just moved the type declaration out of this busy method declaration.

src/schemas/ProgramDonorSummary/resolvers/summaryStats.ts
```diff
@@ -35,15 +35,21 @@ import { Client } from '@elastic/elasticsearch';
...
   GlobalGqlContext,
   BaseQueryArguments & {
     filters: ProgramDonorSummaryFilter[];
   }
-> = esClient => async (source, args, context): Promise<ProgramDonorSummaryStatsGqlResponse> => {
+>;
+
+const programDonorSummaryStatsResolver: (
+  esClient: Client,
+) => DonorStatsResolverType = esClient => async (
```
joneubank, line 48: Moving the type declaration.

src/schemas/ProgramDonorSummary/resolvers/summaryStats.ts
```diff
@@ -140,6 +148,7 @@ const programDonorSummaryStatsResolver: (
       donorsInvalidWithCurrentDictionary: FilterAggregationResult;
       allFilesCount: NumericAggregationResult;
       filesToQcCount: NumericAggregationResult;
+      lastUpdate?: DateAggregationResult;
```
joneubank, line 151: Low cost add, might use this in a UI where the DCC can review if they want to trigger the donor re-index.

blabadi (approved):

PR #385, opened 2021-03-31
Add Topic Permission checking to Kafka endpoint

joneubank:
**Description of changes**

The `/kafka/:topic` resolver is used for external applications to send messages to the DCC kafka. This PR adds permission checks to the endpoint that will require a scope which contains the topic in the policy name. The pattern checking for this scope is handled by the common validation library [icgc-argo/ego-token-utils](https://github.com/icgc-argo/ego-token-utils), with the new rules added in version [8.1.0: canWriteKafkaTopic](https://github.com/icgc-argo/ego-token-utils/pull/59) .

To achieve this, the AuthendticatedRequestMiddleware was extended to return the `serializedScopes` object (scopes as an array of strings instead of scope objects). This auth middleware was added to the kafka router and used with the validations in ego-token-utils.

Changes to the other routes using the auth middleware were also required to adopt the shape changes. For the `donor-aggregation-api` these changes were minimal, but there are complications added to the `file-storage-api`, requiring several query parameters to be cast as strings. The file storage api is behind a feature flag and not enabled anywhere, and requires additional testing before release.

**Type of Change**

- [ ] Bug
- [ ] Styling
- [x] New Feature

### DIFF
--- a/src/routes/kafka-rest-proxy.ts
+++ b/src/routes/kafka-rest-proxy.ts
@@ -22,61 +22,73 @@ import urlJoin from 'url-join';
 import logger from '../utils/logger';
 import fetch from 'node-fetch';
 import { json } from 'body-parser';
-import express from 'express';
+import express, { Router } from 'express';
+import { EgoClient } from 'services/ego';
+import authenticatedRequestMiddleware, {
+  AuthenticatedRequest,
+} from 'routes/middleware/authenticatedRequestMiddleware';
 import egoTokenUtils from 'utils/egoTokenUtils';
 
-var router = express.Router();
-const apiRoot = KAFKA_REST_PROXY_ROOT;
+const createKafkaRouter = (egoClient: EgoClient): Router => {
+  const router = express.Router();
+  const apiRoot = KAFKA_REST_PROXY_ROOT;
 
-// fetch needs to use the json body parser
-router.use(json());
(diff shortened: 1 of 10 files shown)


src/routes/kafka-rest-proxy.ts
```diff
@@ -22,61 +22,73 @@ import urlJoin from 'url-join';
...
-    return res.status(401).send({
-      message: 'expired token',
+    // Require auth
+    if (!authenticated) {
+      res.status(401).json({ error: 'invalid auth token' });
+      return;
+    }
+
+    // Require kafka topic writing permisison
+    const hasPermission = egoTokenUtils.canWriteKafkaTopic({
```
joneubank, line 54: This line is the entire purpose of this PR. All the rest is just clean up on how to pass JWT details into the route handlers.

src/routes/utils/accessValidations.ts
```diff
@@ -30,11 +30,6 @@ import {
 } from '@icgc-argo/ego-token-utils';
 import { EGO_DACO_POLICY_NAME } from 'config';
 
-export const isDccMember = (scopes: PermissionScopeObj[]) => {
```
joneubank, line 33: we don't need to re-implement this check that is performed by ego-token-utils

Buwujiu (approved):

PR #396, opened 2021-04-30
✨ add primary site to file aggregations

samrichca:
**Description of changes**

- add clinical.donor.primary_site to file aggregations so it can be used in the File Repository (facet panel & a chart)
- change the code that creates the argo arranger ES index, to update the config if the index already exists
- update fake file-centric data (compose/file_centric/sample_file_centric.json)

**Type of Change**

- [ ] Bug
- [x] New Feature


joneubank (approved):
looks good!

PR #404, opened 2021-05-19
✨ add file table tsv download

samrichca:
**Description of changes**

- add a route for downloading a TSV of the file table

UI PR: https://github.com/icgc-argo/platform-ui/pull/1942

**Type of Change**

- [ ] Bug
- [x] New Feature


src/routes/file-centric-tsv/index.ts
```diff
@@ -34,19 +36,35 @@ import scoreManifestTsvSchema from './tsvSchemas/scoreManifest';
...
 }): RequestHandler => {
   return async (req, res) => {
-    const { filter: filterStr, fileName }: { filter?: string; fileName?: string } = req.query;
+    const {
+      columns,
+      filter: filterStr,
+      fileName,
+    }: { columns?: string; filter?: string; fileName?: string } = req.query;
+    
+    if (!columns && !tsvSchema) {
```
joneubank, line 61: 👍

src/routes/file-centric-tsv/index.ts
```diff
@@ -61,13 +79,28 @@ const createDownloadHandler = ({
       esQuery,
       sortField: FILE_METADATA_FIELDS['object_id'],
     });
+
+    const columnsSchema =
+      columns &&
+      JSON.parse(decodeURIComponent(columns)).map(
```
joneubank, line 85: Looks like this works correctly. Do we want to move this code into its own method like `buildTsvSchemaFromColumnList(columns: Array<{header:string; getter:string;})`?

joneubank (approved):

PR #426, opened 2021-06-30
Access Filter for TSV downloads

joneubank:
**Description of changes**

Adds JWT based Auth Filters to TSV download endpoints, controlled by META_ACCESS_FILTER flag in env props.

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/routes/file-centric-tsv/index.ts
+++ b/src/routes/file-centric-tsv/index.ts
@@ -19,18 +19,25 @@
 
 import express, { Response, RequestHandler, Request } from 'express';
 import { Client } from '@elastic/elasticsearch';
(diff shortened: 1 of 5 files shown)


src/routes/file-centric-tsv/index.ts
```diff
@@ -40,34 +47,48 @@ const specialColumnHeaders = {
   fileSize: ['Size'],
 };
 
+function combineSqonFilters(filters: {}[]): {} {
```
Buwujiu, line 50: what's sqon?

joneubank: arranger filter objects

Buwujiu (approved):

PR #434, opened 2021-07-09
donor id filters accept array of values

Buwujiu:
**Description of changes**

Fix donor id filter only accepting 1 value by implementing bool should queries to allow donor id filter accept an array of values. 

**Type of Change**

- [x] Bug
- [ ] New Feature

### DIFF
--- a/src/schemas/ProgramDonorSummary/resolvers/summaryEntries.ts
+++ b/src/schemas/ProgramDonorSummary/resolvers/summaryEntries.ts
@@ -71,9 +71,14 @@ const programDonorSummaryEntriesResolver: (
 
     if (field === EsDonorDocumentField.combinedDonorId && filter.values.length > 0) {
       // use wildcard query for donor_id and submitter_donor_id partial match
-      const regex = `*${filter.values[0].toLowerCase()}*`;
-      const wildcardQuery = esb.wildcardQuery(EsDonorDocumentField.combinedDonorId, regex);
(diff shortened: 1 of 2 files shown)


src/schemas/ProgramDonorSummary/resolvers/summaryEntries.ts
```diff
@@ -71,9 +71,14 @@ const programDonorSummaryEntriesResolver: (
...
-      const regex = `*${filter.values[0].toLowerCase()}*`;
-      const wildcardQuery = esb.wildcardQuery(EsDonorDocumentField.combinedDonorId, regex);
-      queries.push( wildcardQuery );
+      const donorIdQueries: Query[] =[];
+      for (const value of filter.values) {
+        const regex = `*${value.toLowerCase()}*`;
+        const wildcardQuery = esb.wildcardQuery(EsDonorDocumentField.combinedDonorId, regex);
+        donorIdQueries.push(wildcardQuery);
+      }
+      const boolQuery = esb.boolQuery().should(donorIdQueries);
```
joneubank, line 80: I believe we intend this to be a must() instead of a should() , since we are not searching for based on matching score, we are filtering just for things that match.

joneubank (approved):
Based on further sources, should() can be used as OR in a bool query, we'll try it this way
https://www.elastic.co/blog/lost-in-translation-boolean-operations-and-filters-in-the-bool-query

PR #460, opened 2021-08-11
Merge donor dashboard api for extending filters

Buwujiu:
**Description of changes**

- Merged `programDonorSummaryStats` and `programDonorSummaryEntries` api into one api `programDonorSummary`, therefor filters can be applied onto all donor dashboard apis. 
- This PR does not remove old apis, will remove them until UI is connected to new api. 

**Type of Change**

- [ ] Bug
- [ ] New Feature


src/schemas/ProgramDonorSummary/gqlTypeDefs.ts
```diff
@@ -120,6 +120,12 @@ export default gql`
     order: SortOrder
   }
 
+  type DonorSummaryEntryAndStats {
+    entry: [DonorSummaryEntry!]
```
joneubank, line 124: lets use a plural name since this is an array type: `entries`

src/schemas/ProgramDonorSummary/gqlTypeDefs.ts
```diff
@@ -289,7 +295,7 @@ export default gql`
     """
     Paginated list of donor data summary given a program
     """
-    programDonorSummaryEntries(
+    programDonorSummaryEntryAndStats(
```
joneubank, line 298: We can simplify the name now `programDonorSummary`

joneubank (changes requested):
Great work, you had the right idea. I left some comment on GQL Schema Type name changes.

Can we leave the existing queries for DonorSummaryEntries (and Stats) and instead just add this new combined query in addition? This way no UI functionality will break. We can remove the duplicated query types after updating the UI.

src/schemas/ProgramDonorSummary/gqlTypeDefs.ts
```diff
@@ -120,6 +120,12 @@ export default gql`
     order: SortOrder
   }
 
+  type DonorSummaryEntryAndStats {
```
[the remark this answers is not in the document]
Buwujiu: 👍

PR #483, opened 2021-09-08
Add index files to score downloads

joneubank:
**Description of changes**
The Index Files associated with a file are included in the documents in the file_centric index. This PR takes advantage of this to include those files in the `/entries` requests from the score client. The end result is that a request for a file download will now also send the index file to the score client for download

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/routes/file-storage-api/handlers/entitiesHandler.ts
+++ b/src/routes/file-storage-api/handlers/entitiesHandler.ts
@@ -149,7 +149,7 @@ const createEntitiesHandler = ({ esClient }: { esClient: Client }): Handler => {
           .boolQuery()
           // TODO: All of the `as string` casting in this section was added to silence the typescript compiler, it needs to be tested
           // the solution is likely in the types being applied to the Request object, such that it doesnt know if the query params are string or parsed into arrays
-          .should([
+          .must([
             parsedRequestQuery.id
               ? esb.termsQuery(FILE_METADATA_FIELDS['object_id'], parsedRequestQuery.id as string)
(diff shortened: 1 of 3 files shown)


src/routes/file-storage-api/handlers/entitiesHandler.ts
```diff
@@ -149,10 +149,16 @@ const createEntitiesHandler = ({ esClient }: { esClient: Client }): Handler => {
           .boolQuery()
           // TODO: All of the `as string` casting in this section was added to silence the typescript compiler, it needs to be tested
           // the solution is likely in the types being applied to the Request object, such that it doesnt know if the query params are string or parsed into arrays
-          .must([
+          .should([
```
Buwujiu, line 152: why change `must` to `should`?

joneubank: Wasn't supposed to. Thought I changed it back. Nice catch.

joneubank: I'm going to review this again tomorrow... I think the extra query I'm adding is not firing in my test cases, and would break if that filter was provided. I think this extra query filter is not needed and should be removed before merging.

joneubank:
@Buwujiu I restored the ES query to its original state since the clause I added was not needed and introduced new breakages. The ES Query in `utils.getEsFileDocumentByObjectId` is what is needed when score requests download by the Index objectId.

Buwujiu (approved):

PR #491, opened 2021-09-29
add option to exclude sample registration downloading

Buwujiu:
**Description of changes**

Adds a new resolver for `/template/all/` endpoint to allow optional downloading for sample registration file.

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/routes/clinical-proxy.ts
+++ b/src/routes/clinical-proxy.ts
@@ -22,10 +22,9 @@ import { CLINICAL_SERVICE_ROOT } from '../config';
 import logger from '../utils/logger';
 import express from 'express';
 import { Request, Response } from 'express';
-const router = express.Router();
-
 import { createProxyMiddleware } from 'http-proxy-middleware';
 
+const router = express.Router();
 // Our specification download service can't use GraphQL because GraphQL specification requires the content-type
 // that it returns be json, and we want to be able to return other content types, such as tab-separated-values,
 // so that the user is automatically prompted to save the file from their browser.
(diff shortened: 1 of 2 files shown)


src/routes/clinical-proxy.ts
```diff
@@ -36,6 +35,25 @@ const handleError = (err: Error, req: Request, res: Response) => {
   return res.status(500).send('Internal Server Error');
 };
 
+router.use(
+  '/template/all/:excludeSampleRegistration',
```
joneubank, line 39: Why is this part of the path instead of a query parameter?

joneubank: Would prefer query param!

joneubank: with a default ;)

src/routes/clinical-proxy.ts
```diff
@@ -36,6 +35,26 @@ const handleError = (err: Error, req: Request, res: Response) => {
   return res.status(500).send('Internal Server Error');
 };
 
+router.use(
+  '/template/all',
+  createProxyMiddleware({
+    target: CLINICAL_SERVICE_ROOT,
+    pathRewrite: (pathName: string, req: Request) => {
+      const param = req.query.excludeSampleRegistration;
+      const exclude = param === undefined ? 'false' : param === 'true';
```
joneubank, line 44:
```
const excludeRegistration = req.query.excludeSampleRegistration === 'true';
```

src/routes/clinical-proxy.ts
```diff
@@ -36,6 +35,26 @@ const handleError = (err: Error, req: Request, res: Response) => {
...
+      return urlJoin('/dictionary/template/all', `?excludeSampleRegistration=${exclude}`);
+    },
+    onError: handleError,
+    changeOrigin: true,
+    onProxyReq(proxyReq, req, res) {
+      const exclude = req.query.excludeSampleRegistration;
+      if (exclude !== 'true' && exclude !== 'false' && exclude !== undefined) {
```
joneubank, line 51:
```
if(excludeRegistration) { ...
```

src/routes/clinical-proxy.ts
joneubank: make this the boolean we want it to be, not a string we need to test against.

joneubank (changes requested):
Thanks for updates. Just some code simplifcation provided.

src/routes/clinical-proxy.ts
Buwujiu: hmm, this condition is to check invalid query param, in case other services like ui sends stuff other than `true` or `false` or `undefined`, gateway can quickly respond with an error without having to reach to clinical.

joneubank (approved):

PR #505, opened 2022-01-10
Add File ID to Extended Mapping

demariadaniel:
**Description of changes**

Adds 'File ID' to file centric extended mapping

**Type of Change**

- [ ] Bug
- [X] New Feature

### DIFF
--- a/src/resources/arranger_es_metadata.json
+++ b/src/resources/arranger_es_metadata.json
@@ -1217,6 +1217,18 @@
             "unit": null,
             "displayValues": {},
             "quickSearchEnabled": false,
+            "field": "file_id",
+            "displayName": "File ID",
+            "active": false,
+            "isArray": false,
+            "rangeStep": 1,
+            "type": "object",
+            "primaryKey": false
+          },
+          {
+            "unit": null,
+            "displayValues": {},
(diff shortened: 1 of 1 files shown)


src/resources/arranger_es_metadata.json
```diff
@@ -1213,6 +1213,18 @@
             "type": "object",
             "primaryKey": false
           },
+          {
+            "unit": null,
+            "displayValues": {},
+            "quickSearchEnabled": false,
+            "field": "file_id",
```
joneubank, line 1220: for reference, this is the mapping used for these fields in ES: https://github.com/icgc-argo/files-service/blob/develop/src/file-centric-index-mapping.json

according to this, it should be nested under file, so `file.file_id`

joneubank (changes requested):
Checked data source mapping and we need to change the field to be `file.file_id`

src/resources/arranger_es_metadata.json
joneubank: nevermind, I misread the nesting. what you have here is correct.

joneubank (approved):
On second read... your changes look like they match the data source.

PR #530, opened 2022-02-24
🔧 Update File Centric ES Mapping to Include `metrics` Object

mistryrn:
**Description of changes**

Adds `metrics` object to the mapping for the file-centric index. From: https://github.com/icgc-argo/files-service/pull/104

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/compose/file_centric/file_mapping.json
+++ b/compose/file_centric/file_mapping.json
@@ -638,18 +638,18 @@
             "type": "long"
           }
         }
-      }
-    },
-    "meta": {
-      "properties": {
-        "study_id": {
-          "type": "keyword"
-        },
-        "release_state": {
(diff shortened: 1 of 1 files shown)


joneubank (approved):
Did you by chance check if the rest of the mapping matched the content in files-service? I would not be surprised to find they are out of sync. Same question re: the meta data repo

mistryrn:
> Did you by chance check if the rest of the mapping matched the content in files-service? I would not be surprised to find they are out of sync. Same question re: the meta data repo

I have not checked that, but I can confirm before merging either of these!

mistryrn:
Mappings have been updated to match the contents of files-service. Thanks for suggesting that!

PR #561, opened 2022-05-30
RNA table filters

Buwujiu:
**Description of changes**

Adding 3 RNA column filters: rna registered samples, rna raw reads, rna alignment status.

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/schemas/ProgramDonorSummary/resolvers/summaryEntriesAndStats.ts
+++ b/src/schemas/ProgramDonorSummary/resolvers/summaryEntriesAndStats.ts
@@ -35,6 +35,7 @@ import {
   workflowStatus,
   DonorSummary,
   ProgramDonorSummaryStats,
+  RnaFilterStatus,
 } from './types';
 import { Client } from '@elastic/elasticsearch';
 import { ELASTICSEARCH_PROGRAM_DONOR_DASHBOARD_INDEX } from 'config';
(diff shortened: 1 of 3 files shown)


src/schemas/ProgramDonorSummary/resolvers/summaryEntriesAndStats.ts
```diff
@@ -259,6 +260,79 @@ const programDonorSummaryEntriesAndStatsResolver: (
...
+            const shouldQueries: Query[] = [];
+            shouldQueries.push(esb.rangeQuery(EsDonorDocumentField.rnaRegisteredNormalSamples).gte(1));
+            shouldQueries.push(esb.rangeQuery(EsDonorDocumentField.rnaRegisteredTumourSamples).gte(1));
+            sampleQueries.push(esb.boolQuery().should(shouldQueries));
+            break;
+        }
+      }
+      const boolQuery = esb.boolQuery().should(sampleQueries);
+      queries.push(boolQuery);
+    }
```
joneubank, line 284: looks like this code is repeated for each field type (`rnaRegisteredSample`, `rnaPublishedNormalAnalysis`, `rnaAlignmentStatus`). probably want to write it out as a function that gets called in each case. easier to maintain changes consistently over each type.

PR #563, opened 2022-06-07
Feat/dep upgrades

ciaranschutte:
**Description of changes**

Upgraded libs and typings

Removed some features not in use. If we require their functionality in the future it will be easier to install a modern library with current implementations than to support deprecated usage.
- removes tracing
- removes cost analysis logging

**Type of Change**

- [ ] Bug
- [ ] New Feature


package.json
```diff
@@ -13,88 +13,92 @@
...
+    "uuid": "^8.3.2"
   },
   "optionalDependencies": {
     "fsevents": "^1.2.9"
   },
   "prettier": {
     "printWidth": 80,
     "trailingComma": "all",
-    "singleQuote": true
+    "singleQuote": true,
```
samrichca, line 101: can we make these the same on the UI repo?

src/services/dataCenterRegistry/index.ts
```diff
@@ -55,7 +55,7 @@ async function fetchDataCenter(code: string): Promise<DataCenter | undefined> {
   })
     .then(async (response: Response) => (await response.json()) as DataCenter)
     .catch(async (response: Response) => {
-      const responseJson = await response.json();
+      const responseJson: any = await response.json();
```
samrichca, line 58: since there's a Response type, is there any chance there's a Response['json']-esque type instead of `any` ?

src/schemas/ProgramDonorSummary/resolvers/summaryEntriesAndStats.ts
```diff
@@ -1062,6 +1063,7 @@ const programDonorSummaryEntriesAndStatsResolver: (
             donorsWithReleasedFilesCount: { doc_count: 0 },
             donorsWithPublishedNormalAndTumourSamples: { doc_count: 0 },
             donorsInvalidWithCurrentDictionary: { doc_count: 0 },
+            donorsWithMatchedTNPair: { doc_count: 0 },
```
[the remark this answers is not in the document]
ciaranschutte: correct

src/services/dataCenterRegistry/index.ts
ciaranschutte: just an object with a string message property. we're not returning responseJson, which would be Promise return type

ciaranschutte: updated

samrichca (approved):

PR #565, opened 2022-06-09
WIP: Feat/2259 core completeness

demariadaniel:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] New Feature

### DIFF
--- a/src/schemas/Clinical/gqlTypeDefs.ts
+++ b/src/schemas/Clinical/gqlTypeDefs.ts
@@ -172,7 +172,7 @@ export default gql`
   """
   input ClinicalInput {
     page: Int!
-    limit: Int!
+    pageSize: Int!
     sort: String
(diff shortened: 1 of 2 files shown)


src/schemas/Clinical/gqlTypeDefs.ts
```diff
@@ -184,7 +184,7 @@ export default gql`
   Collated Clinical Data Query Response
   """
   type ClinicalData {
-    programShortName: String!
+    programShortName: String
```
joneubank, line 187: Is it possible for us to return clinical data with a null programShortName? That seems like it shouldn't be possible...

joneubank (commented):
Looking to understand why programShortName was changed to be nullable in the gql schema, otherwise this is good.

joneubank (approved):

Issue #542, opened 2022-03-23
[API] Fetch an error summary for each file type


kcullion:
- [ ] Fetch an error summary for each file type
- [ ] API to allow us to mark the cells that have an error. 

![image](https://user-images.githubusercontent.com/18424138/159766971-04fb4309-137f-4bfa-af0d-d955f462842c.png)

kcullion:
moving this to awaiting release....made two feedback tickets to track data inconsistencies and UI fixes
https://github.com/icgc-argo/platform-ui/issues/2350
https://github.com/icgc-argo/platform-ui/issues/2355

PR #571, opened 2022-06-17
❗ #542 Submitted Data Error summary

demariadaniel:
**Description of changes**

* Updates ClinicalError response to return errors for different entities
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/icgc-argo/platform-api/542

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/schemas/Clinical/index.ts
+++ b/src/schemas/Clinical/index.ts
@@ -167,7 +167,6 @@ const convertClinicalSubmissionDataToGql = (
 
 type ClinicalEntityData = {
   programShortName: string;
-  totalDocs: number;
   clinicalEntities: ClinicalEntityRecord[];
   completionStats: CompletionStats[];
 };
@@ -208,6 +207,7 @@ type CoreCompletionFields = {
 
 interface ClinicalEntityRecord { 
   entityName: string,
+  totalDocs: number,
   records: EntityRecord[][],
   entityFields: string[],
 };
@@ -230,9 +230,8 @@ const convertClinicalDataToGql = (
   programShortName: string,
   data: any,
 ): ClinicalEntityData => {
-  const { totalDocs, completionStats } = data;
+  const { completionStats } = data;
   const clinicalEntities: ClinicalEntityRecord[] = data.clinicalEntities.map((entity: any) => {
-
     const records: EntityRecord[][] = entity.records.map((record: any) => (
       Object.keys(record)
         .map(key => key && ({ name: key, value: record[key] })
@@ -241,6 +240,7 @@ const convertClinicalDataToGql = (
     const entityData = {
       entityName: entity.entityName,
       entityFields: entity.entityFields,
+      totalDocs: entity.totalDocs,
       records,
     };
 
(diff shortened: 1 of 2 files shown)


src/schemas/Clinical/index.ts
```diff
@@ -233,12 +241,9 @@ type ClinicalErrorRecord = {
...
 }
 
-const convertClinicalDataToGql = (
-  programShortName: string,
-  data: any,
-): ClinicalEntityData => {
-  const clinicalEntities: ClinicalEntityRecord[] = data.clinicalEntities.map((entity: any) => {
-    const records: EntityRecord[][] = entity.records.map((record: any) => (
+const convertClinicalDataToGql = ({programShortName, clinicalEntities}: ClinicalResponseData) => {
+  const clinicalDisplayData: ClinicalEntityRecord[] = clinicalEntities.map((entity: any) => {
```
samrichca, line 245: I think `entity: any` can be `entity: EntityDataRecord`

src/schemas/Clinical/index.ts
```diff
@@ -233,12 +241,9 @@ type ClinicalErrorRecord = {
...
+    const records: EntityDisplayRecord[][] = entity.records.map((record: any) => (
```
samrichca, line 246: Same here - `EntityDataRecord` instead of `any`

src/schemas/Clinical/index.ts
```diff
@@ -593,10 +597,13 @@ const resolvers = {
       context: GlobalGqlContext,
         ) => {
           const { Authorization } = context;
-
-          const donorEntity = parent.clinicalEntities.find(({entityName}) => entityName === CoreClinicalEntities.donor);
-          const donorIds = donorEntity?.completionStats && donorEntity?.completionStats.map(donor => donor.donorId);
-          
+          const entity = parent.clinicalEntities[0];
```
samrichca, line 600: why the switch from matching names (line 597 in the original) to getting the first array item? I assume we only have one item in `clinicalEntities` but will that always be the case...?

demariadaniel: That is a very good catch, I've updated this logic to do a forEach loop over parent.clinicalEntities and find DonorIds for all Entities

The side menus / side-bars need to find which entities have errors so it has to query multiple entities
The submitted data page only displays 1 entity at a time on the page so that's why I was using `clinicalEntities[0]`

src/schemas/Clinical/index.ts
demariadaniel: I've updated these types, please give a second look

src/schemas/Clinical/index.ts
samrichca: nice fix, so this'll work for 1 entity on Submitted Data & N entities in the side menu?

samrichca (approved):

samrichca:
great work, lgtm! 🚀

PR #578, opened 2022-07-12
Feat/dep upgrades elastic search

ciaranschutte:
**Description of changes**

- major ES client 7 - 8 upgade
- related api changes
  - query body key removed 
  - fixed typings 
- minor lib updates

**Type of Change**

- [ ] Bug
- [ ] New Feature


joneubank:
I appreciate the effort that this took, but at this point it is too out of date to be usable. ES upgrade from 7 to 8 brings breaking API changes, so this work will requires quite a bit of coordination to update all our ES interfaces across several services, and of course to the most important tool here which is Arranger. As such this is not on our immediate agenda.

Closing.

PR #585, opened 2022-08-03
↕️ Feat/2265 Entity Data table sort

demariadaniel:
**Description of changes**

Moves Clinical Errors from nested resolver into Clinical Entity query, which allows sorting on API instead of on FE
Changes completionStats type to Float (source of bug)

Relates to: https://github.com/icgc-argo/platform-ui/pull/2358

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/schemas/Clinical/index.ts
+++ b/src/schemas/Clinical/index.ts
@@ -685,32 +685,6 @@ const resolvers = {
       return response ? true : false;
     },
   },
-  // ClinicalData: {
-  //   clinicalErrors: async (
-  //     parent: ClinicalEntityData,
-  //     args: ClinicalVariables,
-  //     context: GlobalGqlContext,
-  //   ) => {
-  //     const { Authorization } = context;
-  //     let donorIds: string[] = [];
-
-  //     parent.clinicalEntities.forEach((entity) =>
-  //       entity.records.forEach((displayRecord: EntityDisplayRecord[]) => {
-  //         const donor = displayRecord.find(({ name }) => name === 'donor_id');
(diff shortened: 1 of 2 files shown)


src/schemas/Clinical/index.ts
```diff
@@ -663,31 +734,6 @@ const resolvers = {
       return response ? true : false;
     },
   },
-  ClinicalData: {
```
demariadaniel, line 666: So the default sort order is to have clinical entity data sorted so that records with clinical errors are sorted to the top. This creates a relationship between clinicalEntities and clinicalErrors, which are separate records from separate DBs. `clinicalData` resolver must now request both `clinicalEntities` and `clinicalErrors` from the service on every query, in order to sort error data to the top.

Currently the FE needs both Clinical Entity & Clinical Error values every time it requests Clinical Data. Pagination limits mean we will never request/return more than 100 entities + 100 errors at any time. Errors are requested by DonorId, using subset of IDs from Clinical Entities.

Ideally we can make the API more single responsibility w/ nested resolvers, but I'm not sure how. @joneubank and I previously separated Entities + Errors by using a nested resolver (Clinical Entities > Clinical Errors). However this creates a parent > child relationship, and Clinical Errors cannot sort the data from Clinical Entities.

Alternately we revert, keep API resolvers single responsibility, and sort by errors on the front end.

demariadaniel: Reverting to Error Sort on FE for now, but I've added a few Types + small improvements through code review 
Saved API Sorting code in the event we align on double requests / nested resolver approach

samrichca (approved):

Issue #610, opened 2022-10-20
Add DNA filters


samrichca:
Related ticket for RNA filters: https://github.com/icgc-argo/platform-api/issues/512
Front-end ticket: https://github.com/icgc-argo/platform-ui/issues/2245
Requirements: https://wiki.oicr.on.ca/display/icgcargotech/Program+Dashboard+Page+Specs#ProgramDashboardPageSpecs-4.1TableDataFields,Filters,&DataSources

- Add a filter for tumor & normal registration. see `percentageTumourAndNormal`
- Add a filter for submitted matched pairs. see `matchedTNPairsDNA`  `donorsWithMatchedTNPair`

Buwujiu:
Tested in QA, found issues, refer to https://github.com/icgc-argo/platform-api/issues/626.

PR #618, opened 2022-10-31
#610: remove graphql-cost-analysis

samrichca:
**Description of changes**

- remove graphql-cost-analysis

**Type of Change**

- [ ] Bug
- [x] New Feature
- [ ] Other: dependency changes


joneubank (commented):
If you're going to do this, lets also remove all the cost of query references in our schemas.

I feel like this needs more of a group discussion before we pull this. It's possible there is some good use to this still.

samrichca:
> If you're going to do this, lets also remove all the cost of query references in our schemas.
> 
> I feel like this needs more of a group discussion before we pull this. It's possible there is some good use to this still.

Agreed! Started a discussion

joneubank (approved):
Thanks!

Issue #639, opened 2022-12-19
programDonorSummary Returns Wrong Info


Buwujiu:
The `programDonorSummary` query returns mismatched info. 

## Issue 1
https://platform-ui.qa.argo.cancercollaboratory.org/submission/program/DEMO-CA/dashboard
There are 11 donors in ES index donor_centric_program_democa_re_48, but the UI is only showing 7.
when querying DEMO-CA through the gateway, 
```
query{
  programDonorSummary(
    programShortName: "DEMO-CA", 
  	first: 20,
  	offset: 0,
  	filters:  [ ] ) {
    entries {
      donorId
      submitterDonorId
      
    }
    stats {
      registeredDonorsCount
    }
  }
}
```
it returns 7 donors only, which should be 11, according to ES.

## Issue 2
https://platform-ui.qa.argo.cancercollaboratory.org/submission/program/DATA-CA/dashboard
DATA-CA dashboard has no donors, but there are 371 in ES donor_centric_program_dataca_re_202,
when querying gateway using the same query as above, it returns 0 donors.

Buwujiu:
tested in QA, working as expected!

Buwujiu:
tested in staging, working as expected!

Buwujiu:
tested in Prod, working as expected!

PR #640, opened 2022-12-21
#639: fix ES scripting

samrichca:
**Description of changes**

- add `value` property to ES builder script
- add env variable for logging ES queries

**Type of Change**

- [x] Bug
- [ ] New Feature


src/schemas/ProgramDonorSummary/resolvers/summaryEntriesAndStats.ts
```diff
@@ -931,6 +935,10 @@ const programDonorSummaryEntriesAndStatsResolver: (esClient: Client) => DonorEnt
 			};
 		};
 
+		if (ENABLE_ELASTICSEARCH_LOGGING) {
```
joneubank, line 938: 🙌

joneubank (approved):
nice

joneubank:
This ES logging is helpful but its incomplete... Just took a look through the code base, we have several other places where esClient is sending a request with a query.

We should probably go through them all and add the logging.

We should also add notes in the Readme about how to enable ES Query logging for troubleshooting.

PR #647, opened 2023-03-20
Accessing Program Service HTTP Public Endpoint 

daniel-cy-lu:
**Description of changes**

Connecting to Program Service's HTTP endpoints by creating a new resolver and error handler. The HTTP endpoints are public. The original endpoints (GRPC) needs authentication. The system will check whether the query contains info that are all public. If that's the case, automatically, the HTTP endpoints is used. And this could be done without auth.

Vice versa, if a query contain private info, then GRPC endpoints are used and this would require auth. 

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/compose/docker-compose.yml
+++ b/compose/docker-compose.yml
@@ -34,7 +34,7 @@ services:
     depends_on:
       - program-db
     environment:
-      # SPRING_PROFILES_ACTIVE: 'auth'
+      SPRING_PROFILES_ACTIVE: 'auth'
       SPRING_DATASOURCE_URL: 'jdbc:postgresql://program-db:5432/program-db?stringtype=unspecified'
       SPRING_DATASOURCE_USERNAME: postgres
       SPRING_DATASOURCE_PASSWORD: password
--- a/src/config.ts
+++ b/src/config.ts
@@ -68,6 +68,7 @@ export const EGO_PUBLIC_KEY =
 
 // External service root
 export const PROGRAM_SERVICE_ROOT = process.env.PROGRAM_SERVICE_ROOT || 'localhost:50052';
+export const PROGRAM_SERVICE_HTTP_ROOT = process.env.PROGRAM_SERVICE_HTTP_ROOT || 'localhost:8083';
 export const CLINICAL_SERVICE_ROOT = process.env.CLINICAL_SERVICE_ROOT || 'http://localhost:3000';
 export const KAFKA_REST_PROXY_ROOT = process.env.KAFKA_REST_PROXY_ROOT || 'http://localhost:8085';
 export const DONOR_AGGREGATOR_API_ROOT =
--- a/src/schemas/Program/index.js
+++ b/src/schemas/Program/index.js
@@ -251,11 +251,22 @@ const resolveProgramList = async (egoToken) => {
 
 const resolveSingleProgram = async (egoToken, programShortName) => {
 	const response = await programService.getProgram(programShortName, egoToken);
-
 	const programDetails = get(response, 'program');
-	console.log('full response', programDetails);
 	return response ? convertGrpcProgramToGql(programDetails) : null;
 };
+
+const resolveHTTPProgram = async (programShortName) => {
+	const response = await programService.getProgramPublicFields(programShortName);
+	return response;
+};
+
+const programServicePrivateFields = [
+	'commitmentDonors',
+	'submittedDonors',
+	'genomicDonors',
+	'membershipType',
+	'users',
+];
 
 const resolvers = {
 	ProgramOptions: {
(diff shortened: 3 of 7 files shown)


compose/docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
     depends_on:
       - program-db
     environment:
-      # SPRING_PROFILES_ACTIVE: 'auth'
+      SPRING_PROFILES_ACTIVE: 'auth'
       SPRING_DATASOURCE_URL: 'jdbc:postgresql://program-db:5432/program-db?stringtype=unspecified'
```
daniel-cy-lu, line 38: GRPC (private) endpoints check for auth

src/config.ts
```diff
@@ -68,6 +68,7 @@ export const EGO_PUBLIC_KEY =
 
 // External service root
 export const PROGRAM_SERVICE_ROOT = process.env.PROGRAM_SERVICE_ROOT || 'localhost:50052';
+export const PROGRAM_SERVICE_HTTP_ROOT = process.env.PROGRAM_SERVICE_HTTP_ROOT || 'localhost:8083';
 export const CLINICAL_SERVICE_ROOT = process.env.CLINICAL_SERVICE_ROOT || 'http://localhost:3000';
```
daniel-cy-lu, line 72: Add a new root for the public http endpoints

src/schemas/Program/index.js
```diff
@@ -255,6 +255,19 @@ const resolveSingleProgram = async (egoToken, programShortName) => {
 	return response ? convertGrpcProgramToGql(programDetails) : null;
 };
 
+const resolveHTTPProgram = async (programShortName) => {
+	const response = await programService.getProgramPublicFields(programShortName);
+	return response;
+};
+
```
daniel-cy-lu, line 262: Add a http public resolver in schemas. The response fetch the right information in Playground. 
Screenshot 2023-03-22 at 10 41 54 AM
Screenshot 2023-03-22 at 10 42 07 AM

src/schemas/Program/index.js
```diff
@@ -255,6 +255,19 @@ const resolveSingleProgram = async (egoToken, programShortName) => {
...
+const programServicePrivateFields = [
+	'commitmentDonors',
+	'submittedDonors',
+	'genomicDonors',
+	'membershipType',
+	'users',
+];
+
```
daniel-cy-lu, line 270: Create a variable for the private fields. More fields could be added in the future if needed.

src/schemas/Program/index.js
```diff
@@ -303,10 +316,35 @@ const resolvers = {
 	},
 	Query: {
 		program: async (obj, args, context, info) => {
+			const requestedFields = info.fieldNodes.flatMap((fieldNode) =>
+				fieldNode.selectionSet.selections.map((selection) => selection.name.value),
+			);
 			const { egoToken } = context;
```
daniel-cy-lu, line 322: `requestedFields` returns an array of strings of the fields requested in the query, for example,` ['name', 'shortName', 'description']`. This is used to match to the private field variable above, `programServicePrivateFields` to determine whether the query is requesting private or public fields. And which of the resolvers to used, `resolveSingleProgram` (private GRPC) or `resolveHTTPProgram` (public HTTP).

src/services/programService/grpcClient.js
```diff
@@ -0,0 +1,354 @@
...
+
+import { get } from 'lodash';
+import grpc from 'grpc';
+import * as loader from '@grpc/proto-loader';
+import protoPath from '@icgc-argo/program-service-proto';
+
+import { PROGRAM_SERVICE_ROOT } from '../../config';
+import { getAuthMeta, wrapValue, withRetries, defaultPromiseCallback } from '../../utils/grpcUtils';
+import retry from 'retry';
+
```
daniel-cy-lu, line 33: This is a refactoring process. In the `programService` folder, we have now an index that receives all the export clients from GRPC and HTTP client files.

src/schemas/Program/index.js
```diff
@@ -303,10 +316,35 @@ const resolvers = {
...
+		// 	context: GlobalGqlContext,
+		// ) => {
+		// 	const { Authorization } = context;
+		// 	const { shortName } = args;
+
+		// 	const response = await clinicalService.getRegistrationData(shortName, Authorization);
+		// 	return convertRegistrationDataToGql(shortName, {
+		// 		registration: response,
+		// 	});
+		// }
```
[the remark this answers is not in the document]
daniel-cy-lu: oh thanks Dan, I don't know why I am staring at it and even put comment here on github and still didn't realize that I need to remove it.. haha It's removed!

demariadaniel (approved):
Great job, lgtm! 🚀

daniel-cy-lu:
Thanks for the review Dan!

src/schemas/Program/index.js
```diff
@@ -303,10 +316,21 @@ const resolvers = {
...
 			const { shortName } = args;
-			return resolveSingleProgram(egoToken, shortName);
+
+			//create a condition to determine using either public or private endpoint
+			const hasPrivateField = requestedFields.filter((requestedField) =>
+				programServicePrivateFields.indexOf(requestedField) != -1 ? true : false,
+			);
+			return hasPrivateField.length > 0
+				? resolveSingleProgram(egoToken, shortName)
+				: resolveHTTPProgram(shortName);
```
ciaranschutte, line 331: this works great - I think we can improve the syntax a little with some of the modern JS api methods.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

```javascript
const hasPrivateField = requestedFields.some((requestedField) => programServicePrivatesFields.includes(requestedField))

return hasPrivateField ? resolveSingleProgram(egoToken, shortName): resolveHTTPProgram(shortName);
```

Why is this better?
`filter` is most often used to filter down elements into a subset. But the only reason for the subset here if to check if there are any elements (`hasPrivateField.length > 0`)

`some` will also early return, so if a condition is met on the 1st element of an array, the condition is fulfilled,instead of iterating the entire array. This works because `hasPrivateField.length > 0` we're not checking if `hasPrivateField.length  = 4` or `hasPrivateField.length < 12`, it just needs to be greater than zero.

Why not use `indexOf`?
we didn't always have an `Array.includes` function, so people would check if an index exists, then return a boolean based on that. Thankfully we do have `Array.includes` now! 
use `includes` when you want a boolean, use `indexOf` when you need to use the actual index of the element.

The nice thing is we don't need the ternary anymore too. Whenever you see a ternary that is `condition ? true : false` there's often a function that will return a boolean.

PR #663, opened 2023-06-12
Update Clinical Data/Error/Search to use POST

joneubank:
**Description of changes**

Updates Clinical API based on breaking changes here: https://github.com/icgc-argo/argo-clinical/pull/980

The clinical data/error/search APIs all use POST method now, with several query params moved into the request body.

This also makes the filter pagination properties useful in gql, adding them to the request to clinical.

This has been tested vs the clinical API locally but not in context in the UI. That will require some further trials after dev deploy.

**Type of Change**

- [ ] Bug
- [x] New Feature

### DIFF
--- a/src/services/clinical/index.js
+++ b/src/services/clinical/index.js
@@ -24,6 +24,8 @@ import urlJoin from 'url-join';
 import { CLINICAL_SERVICE_ROOT } from '../../config';
 import { restErrorResponseHandler } from '../../utils/restUtils';
 
+import { buildClinicalInputQueryString } from './utils';
+
 const getRegistrationData = async (programShortName, Authorization) => {
 	const url = `${CLINICAL_SERVICE_ROOT}/submission/program/${programShortName}/registration`;
 	const response = await fetch(url, {
@@ -66,7 +68,7 @@ const clearRegistrationData = async (programShortName, registrationId, Authoriza
 		},
 	)
 		.then(restErrorResponseHandler)
-		.then((response) => true);
+		.then((_response) => true);
 	return response;
 };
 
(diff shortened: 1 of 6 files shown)


src/services/clinical/index.js
```diff
@@ -137,58 +139,80 @@ const getClinicalSubmissionData = async (programShortName, Authorization) => {
...
+			entityTypes,
+			donorIds: donorIds?.map((id) => Number(id)) || [],
+			submitterDonorIds,
+		},
 	})
 		.then(restErrorResponseHandler)
+		.then((response) => {
+			console.log('hello', response);
+			return response;
+		})
```
demariadaniel, line 166: Testing?

src/services/clinical/index.js
```diff
@@ -137,58 +139,80 @@ const getClinicalSubmissionData = async (programShortName, Authorization) => {
...
 	);
 
 	const response = await fetch(url, {
-		method: 'get',
+		method: 'post',
 		headers: { Authorization },
+		body: {
+			completionState,
+			entityTypes,
+			donorIds: donorIds?.map((id) => Number(id)) || [],
```
demariadaniel, line 158: This line I would review b/c there's no handling for where `Number(id) = NaN`
I would want to make sure front end / clinical validation and filtering are in place

demariadaniel (approved):
Looking great, just minor cleanup and a question about some falsy value checking

src/services/clinical/index.js
joneubank: lol so amateur, i apologize.

src/services/clinical/index.js
joneubank: Good point. I think in this case nothing will break since it will just not find any donors that match ID=NaN. I'll add a filter.

src/services/clinical/index.js
demariadaniel: nbd

src/services/clinical/index.js
demariadaniel: Hmm good question but I'm not sure if we should throw out the whole request for one bad id

For the Search Bar and Filter Modal, if you enter one valid DonorId and one invalid ID, you still receive the results for the correct ID, and the invalid IDs are just ignored. These use a different query (the search query) but I think the same principle should probably apply, especially if a user submits this kind of search

demariadaniel: Error messages for invalid IDs in the clinical response and on the UI could point out bad user input

joneubank: So the GQL Schema will enforce the filter be an array of numbers, making my conversion redundant... I added in a filter for NaN's so we only send valid numeric IDs.

demariadaniel (approved):

PR #677, opened 2023-10-09
Feat/674 rest api integration

daniel-cy-lu:
**Create clients using HTTP to retrieve data from Program Service endpoints. The next step would be to use these new HTTP in place of the existing GRPC clients**

Create clients for getPrivateProgram, listPrivatePrograms, and getPublicProgram. Tested in localhost Playground and all three clients successfully retrieved data from Program Service.

**Type of Change**

- [ ] Bug
- [ ] New Feature
- [x] Tool Update

-  there are three services that are now migrated to http:
    1.  single private program
    2.  private program list
    3.  single public program
    4.  joinProgramInvite (private)
   
- test code for each case:
    1. 
``` 
{
    program(shortName: "ALEXIS-CA") {
    shortName
    description
    name
    commitmentDonors
    submittedDonors
    genomicDonors
    website
    institutions
    countries
    regions
    cancerTypes
    primarySites
    membershipType
  }
}
```
2.
```
{
  programs {
    shortName
    description
    name
    commitmentDonors
    submittedDonors
    genomicDonors
    website
    institutions
    countries
    regions
    cancerTypes
    primarySites
    membershipType

# users is not implement in this PR. listUsers is done in the next PR.
    # users {
    #   email
    #   firstName
    #   lastName
    #   role
    #   isDacoApproved
    #   inviteStatus
    #   inviteAcceptedAt
    # }
  }
}
```
3.
```
{
  program(shortName: "DEMO-CA") {
    shortName
    description
    name
    website
    institutions
    countries
    regions
    cancerTypes
    primarySites
  }
}
```
4.
```
{
  joinProgramInvite(id: "503b60a2-ea97-420f-9a48-8b93b9980d5d") {
    id
    createdAt
    expiresAt
    acceptedAt
    program {
      shortName
      description
      name
      website
      institutions
      countries
      cancerTypes
      primarySites
      commitmentDonors
      submittedDonors
      genomicDonors
    }
    user {
      email
      # first and last name has to be updated in Program Service. Right now they return null
      # firstName
      # lastName
      role
    }
    emailSent
    status
  }
}
```

Note: use bearer authorization for the private queries. The difference 1 and 3 is 3 doesn't have access to private fields such as commitmentDonors, submittedDonors etc. The following is an example of bearer authorization:
```
{"Authorization":"Bearer ${jwt token}"}
```


daniel-cy-lu:
@joneubank does the new error handler apply to all the other http fetchs?

src/schemas/Program/index.js
```diff
@@ -212,7 +212,7 @@ const typeDefs = gql`
 `;
 
 /* =========
-    Convert GRPC Response to GQL output
+HTTP resolvers
  * ========= */
 const getIsoDate = (time) => (time ? new Date(parseInt(time) * 1000).toISOString() : null);
```
daniel-cy-lu, line 217: will remove the rest of GRPC parts later. Need to keep this to prevent errors when testing the existing endpoints.

src/schemas/Program/index.js
```diff
@@ -242,33 +242,19 @@ const convertGrpcUserToGql = (userDetails) => ({
...
-	countries: program.programCountries?.map((country) => country.name) || [],
-	regions: program.processingRegions?.map((region) => region.name) || [],
-	cancerTypes: program.programCancers?.map((cancer) => cancer.name) || [],
-	primarySites: program.programPrimarySites?.map((primarySite) => primarySite.name) || [],
-});
-
-const resolveProgramList = async (egoToken) => {
-	const response = await programService.listPrograms(egoToken);
-	const programs = get(response, 'programs', []);
-	return programs.map((program) => convertGrpcProgramToGql(program));
```
daniel-cy-lu, line 260: moving format handlers into the http fetch file. (maybe refactor the formatter later on)

src/schemas/Program/index.js
```diff
@@ -339,19 +325,18 @@ const resolvers = {
 			);
 
 			return hasPrivateField
-				? resolveSingleProgram(egoToken, shortName)
-				: resolveHTTPProgram(shortName);
+				? resolvePrivateSingleProgram(egoToken, shortName)
+				: resolvePublicSingleProgram(shortName);
 		},
```
daniel-cy-lu, line 330: updating the names to exclude "http", because they are all http now. And include private or public.

src/services/programService/httpClient.js
```diff
@@ -27,14 +27,108 @@ import fetch from 'node-fetch';
...
+//private fields
+export const listPrivatePrograms = async (jwt = null) => {
+	const url = `${PROGRAM_SERVICE_HTTP_ROOT}/programs`;
+	return await fetch(url, {
+		method: 'get',
+		headers: {
+			Authorization: `Bearer ${jwt}`,
+		},
+	})
+		.then(restErrorResponseHandler)
```
daniel-cy-lu, line 73: using this new error handler, `restErrorResponseHandler` after git merge develop.

src/services/programService/httpClient.js
```diff
@@ -27,14 +27,108 @@ import fetch from 'node-fetch';
...
+
+export const getPrivateProgram = async (jwt = null, programShortName) => {
+	const url = `${PROGRAM_SERVICE_HTTP_ROOT}/programs/${programShortName}`;
+	return await fetch(url, {
 		method: 'get',
+		headers: {
+			Authorization: `Bearer ${jwt}`,
+		},
 	})
 		.then(restErrorResponseHandler)
```
daniel-cy-lu, line 93: new error handler in all fetches.

src/services/programService/httpClient.js
```diff
@@ -27,14 +27,116 @@ import fetch from 'node-fetch';
...
+			cancerTypes: invitation.program.programCancers,
+			primarySite: invitation.program.programPrimarySites,
+		},
+	};
+
+	delete formattedObj.program.programInstitutions;
+	delete formattedObj.program.programCountries;
+	delete formattedObj.program.programCancers;
+	delete formattedObj.program.programPrimarySites;
+	return formattedObj;
```
daniel-cy-lu, line 69: remove properties that has been formatted, ex `programInstitutions` is formatted in `institution`, so `programInstitution` is removed.

src/services/programService/httpClient.js
```diff
@@ -27,14 +27,116 @@ import fetch from 'node-fetch';
...
+};
+
+//private fields
+export const listPrivatePrograms = async (jwt = null) => {
+	const url = `${PROGRAM_SERVICE_HTTP_ROOT}/programs`;
```
joneubank, line 74: I would recommend using `url-join` to build URLs, like here: https://github.com/icgc-argo/platform-api/blob/develop/src/services/clinical/api.ts#L42

This can fix some bugs that occur when the service domain is incorrectly formatted (sometimes has a trailing slash, sometimes doesnt... with urlJoin we don't care which).

src/services/programService/httpClient.js
joneubank: I don't understand this comment.

joneubank (changes requested):
This looks good overall. The changes I'm requesting are tangential to your implementation. Specifically, I would like you to:
- [ ] use `url-join` when defining URLs
- [ ] use the `logger` instead of `console.log`

src/services/programService/httpClient.js
daniel-cy-lu: You rewrote the error handler, I git pull in develop and is using the new handler you wrote.

src/services/programService/httpClient.js
daniel-cy-lu: urljoin is implemented. Thanks for showing me this tool! I read the benefits to prevent errors using it!

src/services/programService/httpClient.js
```diff
@@ -27,14 +27,116 @@ import fetch from 'node-fetch';
...
+		});
 };
 
-export { getProgramPublicFields };
+export const getPrivateProgram = async (jwt = null, programShortName) => {
+	const url = `${PROGRAM_SERVICE_HTTP_ROOT}/programs/${programShortName}`;
+	return await fetch(url, {
+		method: 'get',
+		headers: {
+			Authorization: `Bearer ${jwt}`,
```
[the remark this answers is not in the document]
daniel-cy-lu: add a util.js to the folder where I added this function, authorizationHeader. And replaced the string template in the fetch header with authorizationHeader(jwt). manually tested all endpoints and they all work!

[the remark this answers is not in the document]
joneubank: If you are adding this as a new file can you please use typescript and declare the types for the function arguments?

As a general rule, files named `utils` are very ambiguous for what they do and they end up as a dumping ground of random functionality. I would recommend making a dedicated file just for this function in `/src/utils/authorizationHeader.ts` where this function is the default export. That way other files around the application can find and use it.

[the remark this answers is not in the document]
daniel-cy-lu: Got it! Jon, thanks for explaining our convention of using the utils. I added a new folder and within it, added the authorizationHeader.ts. I place the function in the file and added string type to the argument. Then export default this function.

src/services/programService/httpClient.js
```diff
@@ -27,14 +27,116 @@ import fetch from 'node-fetch';
...
+		},
+	})
+		.then(restErrorResponseHandler)
+		.then((response) => response.json())
+		.then((data) => {
+			if (data) {
+				return formatPrivateProgram(data);
+			} else {
+				console.log('Error: no data is returned from /program/{shortName}');
```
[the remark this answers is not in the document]
daniel-cy-lu: Updated console logs to loggers. Also updated message related to the if else condition for future devs to know where the error is coming from.

src/services/programService/httpClient.js
```diff
@@ -104,18 +112,18 @@ export const getPrivateProgram = async (jwt = null, programShortName) => {
...
-				throw new Error('Unable to retrieve program data.');
+				logger.error('Error: no data is returned from /program/{shortName}');
+				throw new Error('No data is returned from /program/{shortName}');
 			}
 		});
 };
 
 export const getJoinProgramInvite = async (jwt = null, id) => {
-	const url = `${PROGRAM_SERVICE_HTTP_ROOT}/programs/joinProgramInvite/${id}`;
+	const url = urljoin(PROGRAM_SERVICE_HTTP_ROOT, `/programs/joinProgramInvite/${id}`);
```
joneubank, line 122: This is good, you could also simplify to:
```js
const url = urljoin(PROGRAM_SERVICE_HTTP_ROOT, `programs/joinProgramInvite`, id);
```

joneubank (approved):
Looks great! Thanks!

src/services/programService/utils/authorizationHeader.ts
```diff
@@ -0,0 +1,28 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * This file dynamically generates a gRPC client from Ego.proto.
+ * The content of Ego.proto is copied directly from: https://github.com/icgc-argo/argo-proto/blob/4e2aeda59eb48b7af20b462aef2f04ef5d0d6e7c/ProgramService.proto
+ */
```
joneubank, line 23: I think you copied in this comment by mistake, it can be removed.

daniel-cy-lu:
@joneubank Thanks Jon!! SO MUCH!

src/services/programService/utils/authorizationHeader.ts
```diff
@@ -0,0 +1,28 @@
...
+
+// Create a function that return this repetitive string template in the fecth header
+const authorizationHeader = (jwt: string) => `Bearer ${jwt}`;
```
joneubank, line 26: You should write this comment as a JSDoc:

```ts
/**
 * Format JWT as Bearer Token for Authorization header.
 * @param {string} jwt 
 * @returns {string}
 * 
 * @example
 * ```
 * await fetch(url, {
 * 	method: 'get',
 * 	headers: {
 * 		Authorization: authorizationHeader(jwt),
 * 	},
 * })
 * ```
 */
const authorizationHeader = (jwt: string) => `Bearer ${jwt}`;

```

joneubank: JS Doc Reference: https://jsdoc.app/about-getting-started

Also, you don't need to specify the variable types in JSDoc when using TypeScript, since we trust the TS types not the types in the docs.

Buwujiu:
@daniel-cy-lu pls make a new ticekt for this PR.

daniel-cy-lu:
@Buwujiu ticket #702 is created and attached for this PR. The PR is unattached to the epic.

PR #680, opened 2023-10-13
🐞  TS Util Issues

demariadaniel:
**Description of changes**

**Type of Change**

- [x] Bug
- [ ] New Feature

### DIFF
--- a/src/schemas/Helpdesk/jiraRequests.ts
+++ b/src/schemas/Helpdesk/jiraRequests.ts
@@ -154,12 +154,12 @@ export const createJiraClient = async (): Promise<JiraClient> => {
 		type ResponseBody = {
 			values: Array<OrganizationCustomer>;
 		};
-		const response = await fetch(url, {
+		const response = (await fetch(url, {
 			method: 'get',
 			headers: requestHeaders,
 		})
 			.then(restErrorResponseHandler)
-			.then((response) => response.json() as ResponseBody);
+			.then((response) => response.json())) as ResponseBody;
 
 		const id = response.values.find(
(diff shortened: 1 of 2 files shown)


src/schemas/Helpdesk/jiraRequests.ts
```diff
@@ -154,12 +154,12 @@ export const createJiraClient = async (): Promise<JiraClient> => {
...
 			values: Array<OrganizationCustomer>;
 		};
-		const response = await fetch(url, {
+		const response = (await fetch(url, {
 			method: 'get',
 			headers: requestHeaders,
 		})
 			.then(restErrorResponseHandler)
-			.then((response) => response.json() as ResponseBody);
+			.then((response) => response.json())) as ResponseBody;
```
ciaranschutte, line 162: https://github.com/EUCANCan-Federated-Data-Portal/variant-indexer/blob/main/src/external/song/SongClient.ts#L84

there's a cool way to validate `response.json()` results similar to something Jon has wrote here. Where you can use zod to verify the response.

I realise using `as` is how it's done already - just sharing for knowledge sake

ciaranschutte (approved):

src/schemas/Helpdesk/jiraRequests.ts
demariadaniel: `safeParse` ?

ciaranschutte: that function yes, you'd have a zod schema of what you expect the response to be
then use `safeParse` to verify that the response is what you expect
similar to what you'd do if you wrote a typeguard

PR #682, opened 2023-10-13
Replace config type assertions with defaults declaring missing variable

joneubank:
**Description of changes**

Removed type assertions reading process.env variables.

For the variables that are required to have a value (`undefined` is breaking) I have given the config value a default string that will be invalid in that context, the default value will declare the env variable is missing. This should help troubleshooting cases where this env value is missing!

**Type of Change**

- [x] Bug
- [ ] New Feature

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -23,8 +23,9 @@ export const ADVERTISED_HOST = process.env.ADVERTISED_HOST || 'http://localhost:
 
 // Elasticsearch config
 export const ELASTICSEARCH_HOST = process.env.ELASTICSEARCH_HOST || 'http://localhost:9200';
-export const ELASTICSEARCH_VAULT_SECRET_PATH = process.env
-	.ELASTICSEARCH_VAULT_SECRET_PATH as string;
+export const ELASTICSEARCH_VAULT_SECRET_PATH =
+	process.env.ELASTICSEARCH_VAULT_SECRET_PATH ||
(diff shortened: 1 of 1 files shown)


ciaranschutte (approved):

src/config.ts
```diff
@@ -44,28 +45,27 @@ export const ARRANGER_PROJECT_ID = process.env.ARRANGER_PROJECT_ID || 'argo';
...
-export const EGO_VAULT_SECRET_PATH = process.env.EGO_VAULT_SECRET_PATH as string;
+export const EGO_VAULT_SECRET_PATH =
+	process.env.EGO_VAULT_SECRET_PATH || 'Missing env variable EGO_VAULT_SECRET_PATH';
 export const EGO_CLIENT_ID = process.env.EGO_CLIENT_ID;
 export const EGO_CLIENT_SECRET = process.env.EGO_CLIENT_SECRET;
 
 // Ego Credentials for Score Proxy
-export const EGO_VAULT_SCORE_PROXY_SECRET_PATH = process.env
-	.EGO_VAULT_SCORE_PROXY_SECRET_PATH as string;
+export const EGO_VAULT_SCORE_PROXY_SECRET_PATH = process.env.EGO_VAULT_SCORE_PROXY_SECRET_PATH;
```
justincorrigible, line 54: out of curiosity, no fallback string here?

PR #694, opened 2023-10-23
Feat/674 rest api integration v2

daniel-cy-lu:
**Description of changes**

**Type of Change**

- [ ] Bug
- [ ] New Feature


src/services/programService/httpClient.js
```diff
@@ -27,14 +27,292 @@ import fetch from 'node-fetch';
...
+			if (data && Array.isArray(data)) {
+				return formatAndSortMultipleLists(data);
+			} else {
+				console.log('Error: no data is returned from /programs/countries');
+				throw new Error('Unable to retrieve countries data.');
+			}
+		});
+};
+
+export const createProgram = async (program, jwt = null) => {
```
daniel-cy-lu, line 268: @joneubank This is the createProgram function. The formattedProgram is the format in swagger.

demariadaniel:
This PR updates queries to use HTTP for data fetching

GRPC code removal is moved to this PR: https://github.com/icgc-argo/platform-api/pull/714

src/schemas/Program/index.js
```diff
@@ -292,45 +58,39 @@ const resolvers = {
...
 	},
 	Program: {
 		users: async (program, args, context, info) => {
 			const { egoToken } = context;
-			const response = await programService.listUsers(program.shortName, egoToken);
-			const users = response ? get(response, 'userDetails', []).map(convertGrpcUserToGql) : null;
-			return users;
+			const users = await programService.listUsers(egoToken, program.shortName);
+			return users || null;
+		},
```
demariadaniel, line 89: Investigating bug 'cannot return null for non-nullable field ProgramUser.email'

demariadaniel: OK as far as I can tell, several Users exist with `role = null`
This did not match the Type def, so the entire User was returned as `null`, likely causing the above `ProgramUser.email` error (I could not find any records where email was null)

User objects with `role = null` also caused problems for Puneet; she was unable to open an email invite link
Some of the User objects where `role = null` also have `"inviteStatus": "EXPIRED"`, so it might make sense in these cases

For now I am making the field nullable to allow the Query to succeed
But this needs a follow up ticket as it is also the source of bugs, and we should make sure User role is only `null` in acceptable cases

src/schemas/Program/index.js
```diff
@@ -17,251 +17,17 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { ApolloError, gql, UserInputError } from 'apollo-server-express';
+import typeDefs from './gqlTypeDefs';
```
joneubank, line 20: thank you for organizing.

joneubank (approved):

PR #695, opened 2023-10-24
Download Clinical Data by File API Updates

joneubank:
**Description of changes**

- Cleanup text in Arranger mapping for `has_clinical_data` field
- Switch to `object_id` to identify files for the download clinical data by file endpoint
    - Small change greatly simplifies changes required on UI


justincorrigible (approved):
Me like! :shipit:

Issue #693, opened 2023-10-19
Add `dataCenter` to GQL `Program` type


joneubank:
## Detailed Description
All program responses can include details on their assigned DataCenter. This data is being returned for all programs from the Program Service, so this can be added to the GQL type

- [ ] Add `dataCenter` property to the `Program` type in the GQL Schema
- [ ] Ensure that the program resolvers all format and return this data center data correctly

Buwujiu:
Blocked by program service PR merging.

Issue #690, opened 2023-10-19
Remove `regions` from Progrqam GraphQL Schema and Resolvers


joneubank:
Regions are being replaced by data centers in the programs data model. To prep for this we will remove `regions` from the gateway.

This ticket should be done along side https://github.com/icgc-argo/platform-ui/issues/2724 which makes similar changes to the UI. There is a cross dependency on the regions field in the create and update program methods, so changes will need to be deployed to UI and API at the same time.

## Detailed Description
- [ ] Remove `regions` from the program type in the GQL schema
- [ ] Remove `regions` references in the program resolvers

demariadaniel:
Regions need to be replaced by dataCenters
Causing bug in context with Create Program / Update Program operations
cc: @joneubank

PR #704, opened 2023-11-16
🗺️  662 Filter Programs by Region

demariadaniel:
**Description of changes**

Filters programs by requested regions

**Type of Change**

- [ ] Bug
- [ ] New Feature


daniel-cy-lu:
@joneubank @demariadaniel I have ticket #690 . Where I am to remove the "regions" field in the Program Type. I see Dan is using program.regions as part of his filter and this would be broken if I am to remove regions.

demariadaniel:
> @joneubank @demariadaniel I have ticket #690 . Where I am to remove the "regions" field in the Program Type. I see Dan is using program.regions as part of his filter and this would be broken if I am to remove regions.

Thanks for mentioning, I saw that too, makes things confusing

daniel-cy-lu:
@demariadaniel Jon was at my desk looking at this PR. So we won't be using regions going forward. Instead we will be using dataCenter field. So maybe we do the same thing filtering by dataCenter. I have a ticket #693 that I will be adding dataCenter to the Program type. It had a blocker that just got resolved. I will work on it now, so Dan you can modify your filter.

demariadaniel:
> @demariadaniel @joneubank Jon was at my desk looking at this PR. So we won't be using regions going forward. Instead we will be using dataCenter field. So maybe we do the same thing filtering by dataCenter. I have a ticket #693 that I will be adding dataCenter to the Program type. It had a blocker that just got resolved. I will work on it now, so Dan you can modify your filter.

Thank you! I will revert this to a `draft` PR for now. I will follow your changes!

daniel-cy-lu:
@demariadaniel dataCenter is ready. It's a property in the gql Program type.

joneubank (changes requested):
It would be better to modify the existing `programs` query to add an optional filter by datacenter ID. We've definitely added new queries like this before but you can see the amount of noise that generates.

Can we rebuild it as a filter instead of as a new query?

demariadaniel:
> It would be better to modify the existing `programs` query to add an optional filter by datacenter ID. We've definitely added new queries like this before but you can see the amount of noise that generates.
> 
> Can we rebuild it as a filter instead of as a new query?

Yes, good call. Updated in https://github.com/icgc-argo/platform-api/pull/704/commits/821fde1e991a136dfe5724a86e666348fd41c3c5

src/schemas/Program/index.js
```diff
@@ -352,10 +351,19 @@ const resolvers = {
...
-		programs: async (obj, args, context, info) => {
+		programs: async (obj, args, context) => {
 			const { egoToken } = context;
-			return resolvePrivateProgramList(egoToken);
+			const { dataCenter } = args;
+
+			const programs = await resolvePrivateProgramList(egoToken);
+
+			const filteredPrograms = programs.filter(
+				(program) => !dataCenter || program.dataCenter?.shortName === dataCenter,
```
ciaranschutte, line 361: I think you can avoid the filter iterations completely if there's no dataCenter arg?

demariadaniel: Yes! Updated

src/schemas/Program/index.js
```diff
@@ -352,10 +351,19 @@ const resolvers = {
...
+		programs: async (obj, args, context) => {
 			const { egoToken } = context;
-			return resolvePrivateProgramList(egoToken);
+			const { dataCenter } = args;
+
+			const programs = await resolvePrivateProgramList(egoToken);
+
+			const filteredPrograms = !dataCenter
+				? programs
+				: programs.filter((program) => program.dataCenter?.shortName === dataCenter);
```
joneubank, line 362: can you invert the check here? simpler to read with the positive condition:


```ts
			const filteredPrograms = dataCenter
				? programs.filter((program) => program.dataCenter?.shortName === dataCenter)
				: programs;
```

joneubank (changes requested):
Great change, thanks. Minor request here.

src/schemas/Program/index.js
demariadaniel: Np sorry about that, updated

ciaranschutte (approved):
looks good to me

ciaranschutte:
@demariadaniel ah I thought my approval would be enough to merge. looks like you specifically need whoever requested changes approval. @joneubank

joneubank (approved):

Issue #689, opened 2023-10-19
Remove GRPC Clients and Configs


joneubank:
We have removed all GRPC dependencies from the program service and from Ego, so now we want to remove those from the platform API. In particular the code that established the GRPC connection needs to be removed.

There is GRPC code written for Ego and for Program Service - all of it should be removed.

## Detailed Description
- [ ] Remove Program Service GRPC client
- [ ] Remove any code creating GRPC connection to Program Service
- [ ] Remove config variables related to GRPC
- [ ] Remove Ego GRPC client code

PR #714, opened 2024-02-05
✂️ #689 Remove GRPC Client and Config

demariadaniel:
**Description of changes**

Removes all references to GRPC code

https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/platform-api/689

**Type of Change**

- [ ] Bug
- [ ] New Feature


src/schemas/Program/index.js
```diff
@@ -235,33 +242,6 @@ const typeDefs = gql`
 /* =========
 HTTP resolvers
  * ========= */
-const getIsoDate = (time) => (time ? new Date(parseInt(time) * 1000).toISOString() : null);
```
demariadaniel, line 238: See #694, many of these changes are from that branch
https://github.com/icgc-argo/platform-api/pull/694

src/services/programService/httpClient.js
```diff
@@ -18,7 +18,7 @@
  */
 
 /*
- * This file dynamically generates a gRPC client from Ego.proto.
+ * This file dynamically generates an HTTP client from Ego.proto.
  * The content of Ego.proto is copied directly from: https://github.com/icgc-argo/argo-proto/blob/4e2aeda59eb48b7af20b462aef2f04ef5d0d6e7c/ProgramService.proto
  */
```
joneubank, line 23: This comment doesn't apply to the HTTP client, can we remove?

src/services/ego/index.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 /*
- * This file dynamically generates a gRPC client from Ego.proto.
+ * This file dynamically generates an HTTP client from Ego.proto.
  * The content of Ego.proto is copied directly from: https://github.com/overture-stack/ego/blob/develop/src/main/proto/Ego.proto
  */
```
joneubank, line 23: Another comment that means nothing now that we have no grpc, let's remove it.

joneubank (approved):
Couple comments can be cleared out, otherwise great stuff.

src/services/programService/index.js
```diff
@@ -18,33 +18,32 @@
...
-	updateUser: grpc.updateUser,
-	removeUser: grpc.removeUser,
+	inviteUser: http.inviteUser,
+	joinProgram: http.joinProgram,
+	updateUser: http.updateUser,
+	removeUser: http.removeUser,
 
 	getPublicProgram: http.getPublicProgram,
-	listDataCenters: http.listDataCenters,
 };
```
ciaranschutte, line 49: I don't understand this file. Isn't this just the same as exporting the http object `export default http`?
And even then that's all it's doing. There's no mapping to new property names. Just a re-export.

src/services/ego/index.ts
demariadaniel: Yep I had a feeling these could be removed

src/schemas/Program/index.js
```diff
@@ -388,66 +363,78 @@ const resolvers = {
...
 			const { egoToken } = context;
 			const updates = pickBy(get(args, 'updates', {}), (v) => v !== undefined);
-			const shortName = get(args, 'shortName', {});
-
+			const programShortName = get(args, 'shortName', {});
+			const dataCenterShortName = args.updates.dataCenterShortName;
+			delete updates.dataCenterShortName;
+
+			//make a request to get dataCenter data using dataCenterShortName (from query paramenter, updates) and format response
+			const [dataCenterResponse] = await programService.listDataCenters(
```
ciaranschutte, line 379: is the return value of `listDataCenters` an iterable?

src/schemas/Program/index.js
```diff
@@ -388,66 +363,78 @@ const resolvers = {
...
 
 		updateProgram: async (obj, args, context, info) => {
+			// extract information from query parameters
 			const { egoToken } = context;
 			const updates = pickBy(get(args, 'updates', {}), (v) => v !== undefined);
-			const shortName = get(args, 'shortName', {});
-
+			const programShortName = get(args, 'shortName', {});
+			const dataCenterShortName = args.updates.dataCenterShortName;
+			delete updates.dataCenterShortName;
```
ciaranschutte, line 376: could assign the var `updates` above exactly how you want it once, instead of deleting a property.
using `delete` can lead to functionality being hard to follow vs defined once and not mutated

src/services/programService/index.js
demariadaniel: I was questioning this as well, I will investigate

src/schemas/Program/index.js
demariadaniel: Yes these are Daniel's original changes, I will revisit this once the original branch is merged.
One issue is `updates` is not Typed, so I figured `delete` was a simple way to accomplish what's being done here, vs making a method for duplicating an object minus one key

src/services/programService/index.js
joneubank: Agree we can simplify this now. Previously some endpoints were grpc and some were http. Now this isn't needed.

src/schemas/Program/index.js
ciaranschutte: ah gotcha. Yeah there's not a whole lot of typing to assist.
I wouldn't worry about making a method.. there's plenty of lodash util use already, you could just do an `omit`. returns a new object.

i'm not actually sure why pickby is in use either to be honest. Looks like it's just checking for undefined values but I thought that's enforced by the gql endpoints because of it's typing ?

src/schemas/Program/index.js
demariadaniel: Yes it is returning an array or throwing an error

src/services/programService/index.js
demariadaniel: Changed `httpClient` to just `index` and added exports to the file
https://github.com/icgc-argo/platform-api/pull/714/commits/3de29a1768d66dfe6cce44a58673192e757f30e6

src/schemas/Program/index.js
ciaranschutte: Sorry I'm trying to pin down what I'm confused on here. 

it's returning an array but I'm confused why it's returning an array.
It's just destructing the first element, and throwing away any other elements.
Then destructing again on line #383 to just use the same object in the `combinedUpdates` 

Not saying this is a blocker. Just wondering if there's a reason we're getting more data than we need?

src/schemas/Program/index.js
demariadaniel: See how this looks instead? https://github.com/icgc-argo/platform-api/pull/714/commits/8d3f99dae9e966c650a5dc8a2d701714d4d711c6

I checked, there is no use of the Data Center on Platform UI yet, so we should be able to change the types around for now

src/services/programService/index.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 The Ontario Institute for Cancer Research. All rights reserved
+ * Copyright (c) 2022 The Ontario Institute for Cancer Research. All rights reserved
```
justincorrigible, line 2: lol
```suggestion
 * Copyright (c) 2024 The Ontario Institute for Cancer Research. All rights reserved
```

demariadaniel: Ha, that snuck through. Fixed.

ciaranschutte (approved):
code looks good for scope of this ticket 👍

PR #718, opened 2024-02-22
🪰 #1098 Fix Update Program Input

demariadaniel:
**Description of changes**

Update Program receives a DataCenter Id from UI, this updates resolver to work accordingly

https://github.com/icgc-argo/platform-api/issues/693
https://github.com/icgc-argo/roadmap/issues/1098

**Type of Change**

- [ ] Bug
- [ ] New Feature

### DIFF
--- a/src/schemas/Program/index.js
+++ b/src/schemas/Program/index.js
@@ -146,12 +146,12 @@ const resolvers = {
 		updateProgram: async (obj, args, context, info) => {
 			// extract information from query parameters
 			const { egoToken } = context;
-			const { shortName: programShortName, dataCenterShortName } = args;
+			const { shortName: programShortName } = args;
 			const updates = args.updates;
 
 			//make a request to get dataCenter data using dataCenterShortName (from query paramenter, updates) and format response
 			const dataCenterResponse = updates.dataCenter
-				? await programService.listDataCenters(updates.dataCenter, egoToken)
+				? await programService.listDataCenters(undefined, egoToken, updates.dataCenter)
 				: undefined;
 
 			const dataCenter =
@@ -171,7 +171,8 @@ const resolvers = {
 
 			//make a request to the PUT updateProgram endpoint with the formatted payload
 			const response = await programService.updateProgram(combinedUpdates, egoToken);
-			return response === null ? null : programShortName;
+
+			return response === 200 ? programShortName : null;
 		},
 
 		inviteUser: async (obj, args, context, info) => {
--- a/src/services/programService/index.js
+++ b/src/services/programService/index.js
@@ -45,6 +45,9 @@ const formatPrivateProgramList = (programList) => programList.map(formatPrivateP
 const getDataCenterByShortName = (shortName, dataCenterResponse) =>
 	dataCenterResponse.filter((dataCenterObject) => dataCenterObject.shortName === shortName);
 
+const getDataCenterById = (id, dataCenterResponse) =>
+	dataCenterResponse.filter((dataCenterObject) => dataCenterObject.id === id);
+
 const formatJoinProgramInvite = (invitation) => {
 	const formattedObj = {
 		...invitation,
@@ -318,7 +321,7 @@ export const listCountries = async (jwt = null) => {
 		});
 };
 
-export const listDataCenters = async (shortName, jwt) => {
+export const listDataCenters = async (shortName, jwt, id) => {
 	const url = urljoin(PROGRAM_SERVICE_HTTP_ROOT, `/datacenters`);
 	return await fetch(url, {
 		method: 'get',
@@ -330,7 +333,12 @@ export const listDataCenters = async (shortName, jwt) => {
 		.then((response) => response.json())
 		.then((data) => {
 			if (data && Array.isArray(data)) {
-				return shortName ? getDataCenterByShortName(shortName, data) : data;
+				if (shortName) {
+					return getDataCenterByShortName(shortName, data);
+				} else if (id) {
+					return getDataCenterById(id, data);
+				}
+				return data;
 			} else {
 				logger.error('Error: no data or wrong data type is returned from GET /datacenters. Data must be an array.');
 				throw new Error('No data or wrong data type is returned from GET /datacenters. Data must be an array.');
@@ -365,7 +373,7 @@ export const updateProgram = async (programInput, jwt = null) => {
 		body: JSON.stringify(programInput),
 	})
 		.then(restErrorResponseHandler)
-		.then((response) => response.body);
+		.then((response) => response.status);
 };
 
 export const inviteUser = async (userInput, jwt = null) => {


src/services/programService/index.js
```diff
@@ -45,6 +45,9 @@ const formatPrivateProgramList = (programList) => programList.map(formatPrivateP
 const getDataCenterByShortName = (shortName, dataCenterResponse) =>
 	dataCenterResponse.filter((dataCenterObject) => dataCenterObject.shortName === shortName);
 
+const getDataCenterById = (id, dataCenterResponse) =>
+	dataCenterResponse.filter((dataCenterObject) => dataCenterObject.id === id);
```
ciaranschutte, line 49: are we expecting 1 data or many?
from the name I'm assuming a single data center - so `Array.find` might work better

src/services/programService/index.js
```diff
@@ -330,7 +333,12 @@ export const listDataCenters = async (shortName, jwt) => {
 		.then((response) => response.json())
 		.then((data) => {
 			if (data && Array.isArray(data)) {
```
ciaranschutte, line 335: this can just be `Array.isArray(data)` , the isArray method handles undefined vs something like `Array.isArray(data.x)` where `x` will be looked up first and may not exist

src/services/programService/index.js
```diff
@@ -318,7 +321,7 @@ export const listCountries = async (jwt = null) => {
 		});
 };
 
-export const listDataCenters = async (shortName, jwt) => {
+export const listDataCenters = async (shortName, jwt, id) => {
 	const url = urljoin(PROGRAM_SERVICE_HTTP_ROOT, `/datacenters`);
```
ciaranschutte, line 325: `const url = urljoin(PROGRAM_SERVICE_HTTP_ROOT, 'datacenters');`
should be ok as this - url join should handle the slash and there's no need to use a template string because there's no variables being interpolated

src/services/programService/index.js
```diff
@@ -365,7 +373,7 @@ export const updateProgram = async (programInput, jwt = null) => {
 		body: JSON.stringify(programInput),
 	})
 		.then(restErrorResponseHandler)
-		.then((response) => response.body);
+		.then((response) => response.status);
```
ciaranschutte, line 376: not a blocker or huge concern - just overall maybe we should stick with returning body/data/objects from the "service functions" and let the layer above handle the HTTP statuses. Penny for your thoughts as the other functions seem to not return statuses.

src/services/programService/index.js
demariadaniel: Ah, great call, see what you mean, will update

src/services/programService/index.js
```diff
@@ -330,7 +333,12 @@ export const listDataCenters = async (shortName, jwt) => {
 		.then((response) => response.json())
 		.then((data) => {
 			if (data && Array.isArray(data)) {
-				return shortName ? getDataCenterByShortName(shortName, data) : data;
+				if (shortName) {
+					return getDataCenterByShortName(shortName, data);
+				} else if (id) {
+					return getDataCenterById(id, data);
+				}
+				return data;
```
ciaranschutte, line 341: This function `listDataCenters` is confusing to me. 
Sometimes an `id` is passed, sometimes a `shortName` is passed and sometimes we just return the response without worrying about either.

Two suggested approaches:
1) change the function to take an options object. something like:
`const listDataCenters = async ({shortName, jwt, id}) => {..}`
and then there's no need to call this function with `undefined` parameters: https://github.com/icgc-argo/platform-api/pull/718/files#diff-36c4aa96f1bb25a05073048f02c5384b237fd2892af6c33d2d9dd7907d471e8eR154
then add some documentation with the `if/else` return section - it's not clear why we'd just return resp data 

2) my preferred method would be to extract the http request and break this into two functions that have a single responsibility. something like:
```js
const listDataCenters = async (jwt) => {
	const url = urljoin(PROGRAM_SERVICE_HTTP_ROOT, `/datacenters`);
	return await fetch(url, {
		method: 'get',
		headers: {
			Authorization: authorizationHeader(jwt),
		},
	})
		.then((response) => response.json())
};

export const getDataCenterByShortName = (shortName,jwt) => {
	const d = httpreq(jwt)
	return alreadyCreatedFilterFunc(d)
}

export const getDataCenterById = (id,jwt) => {
	const d = httpreq(jwt)
	return alreadyCreatedFilterFunc(d)
}


```
There's already the function logic in this file, it's just fully breaking it into it's own thing.
This way the functions are clear from just the name and are responsible for one thing.
You can reuse the common functionality of the HTTP req.

ciaranschutte (changes requested):
couple of suggestions. let me know if I can help with any

src/services/programService/index.js
demariadaniel: Yes agree, this could have used a second look before opening PR for review, will probably go w/ exported functions, will review

src/services/programService/index.js
demariadaniel: Yes this change is not necessary, program service is returning a pass-through response w/ no body (that we need) from this endpoint so I was thinking 'just look for pass/fail'
which you basically get if there's a request body anyways

src/services/programService/index.js
```diff
@@ -45,6 +45,9 @@ const formatPrivateProgramList = (programList) => programList.map(formatPrivateP
 const getDataCenterByShortName = (shortName, dataCenterResponse) =>
 	dataCenterResponse.filter((dataCenterObject) => dataCenterObject.shortName === shortName);
```
ciaranschutte, line 46: should this be a `find` too?

ciaranschutte (approved):

src/services/programService/index.js
demariadaniel: Yes, updated

ciaranschutte:
I'm unclear if need a get request for both `id` and `shortName`. May make sense if we consistently use one. I'll leave that up to your discretion. I've just made an example with one but the same thing would apply if need both.

Here's the changes I'm suggesting in code: 
https://github.com/icgc-argo/platform-api/pull/719

ciaranschutte (dismissed):
re approving this as the additional `find` was fixed and the other code is working

src/schemas/Program/index.js
```diff
@@ -146,18 +149,19 @@ const resolvers = {
...
+			const { shortName: programShortName } = args;
 			const updates = args.updates;
 
 			//make a request to get dataCenter data using dataCenterShortName (from query paramenter, updates) and format response
 			const dataCenterResponse = updates.dataCenter
-				? await programService.listDataCenters(updates.dataCenter, egoToken)
+				? await programService
+						.listDataCenters(egoToken)
+						.then((data) => programService.getDataCenterById(updates.dataCenter, data))
 				: undefined;
```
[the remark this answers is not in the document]
demariadaniel: Made a minor cleanup to improve readability for now

demariadaniel:
> I'm unclear if need a get request for both `id` and `shortName`. May make sense if we consistently use one. I'll leave that up to your discretion. I've just made an example with one but the same thing would apply if need both.
> 
> Here's the changes I'm suggesting in code: #719

Discussed on Slack -- use of `id` is a by product of shared UI Create/Update Program form
Create Program requires a dataCenter Id, so Update Program also receives a dataCenter id on the API

ciaranschutte (approved):

PR #721, opened 2025-01-23
Endpoint to download clinical data for a filtered query

ciaranschutte:
**Link to Issue**
- icgc-argo/roadmap#1056

**Description of changes**

- adds support for "download all files" from file repo
- uses Arranger for inbuilt permission management
- queries arranger GQL endpoint on self, can't target Arranger api directly as it's not exported from module (using old version of Arranger)
- performant even with 12k files selected, looks like we have internal Arranger code already handling large selections https://github.com/overture-stack/arranger/blob/legacy/modules/mapping-utils/src/resolveHits.js#L38

**Type of Change**

- [ ] Bug
- [ ] New Feature

### DIFF
--- a/src/resources/swagger/index.yaml
+++ b/src/resources/swagger/index.yaml
@@ -102,8 +102,8 @@ paths:
         - Clinical API
       security:
         - bearerAuth: []
-      summary: Download clinical data for all available files from filter.
-      description: Download clinical data for all available files from filter.
+      summary: Download clinical data associated with the donors for all files. A SQON filter can be provided to restrict the data retrieved to the files returned by a specific query.
+      description: Download clinical data associated with the donors for all files. A SQON filter can be provided to restrict the data retrieved to the files returned by a specific query.
       responses:
         '200':
           description: 'A zip file with all program submitted data'
--- a/src/resources/swagger/index.yaml
+++ b/src/resources/swagger/index.yaml
@@ -89,7 +89,7 @@ paths:
         '500':
           description: Internal server error
 
-  '/clinical/api/donors/data-for-all-files':
+  '/clinical/api/donors/data-for-query':
     parameters:
       - name: filter
         description: JSON filter following the SQON spec
--- a/src/resources/swagger/index.yaml
+++ b/src/resources/swagger/index.yaml
@@ -53,7 +53,7 @@ paths:
                 type: string
         '500':
           description: Internal server error
-          
+
   '/clinical/api/donors/data-for-files':
     post:
       tags:
@@ -88,6 +88,35 @@ paths:
           description: The error message indicates something wrong with the request
         '500':
           description: Internal server error
+
+  '/clinical/api/donors/data-for-all-files':
+    parameters:
+      - name: filter
+        description: JSON filter following the SQON spec
+        in: query
+        required: false
+        schema:
+          type: string
+    post:
+      tags:
+        - Clinical API
+      security:
+        - bearerAuth: []
+      summary: Download clinical data for all available files from filter.
+      description: Download clinical data for all available files from filter.
+      responses:
+        '200':
+          description: 'A zip file with all program submitted data'
+          content:
+            application/zip:
+              schema:
+                type: string
+                format: binary
+        '400':
+          description: The error message indicates something wrong with the request
+        '500':
+          description: Internal server error
+
   '/clinical/proxy/program/{programId}/all-clinical-data':
     parameters:
       - name: programId
--- a/src/routes/clinical/arranger.ts
+++ b/src/routes/clinical/arranger.ts
@@ -1,0 +1,95 @@
+import { ADVERTISED_HOST } from 'config';
+import { get } from 'lodash';
+import fetch from 'node-fetch';
+import urljoin from 'url-join';
+
+const url = urljoin(ADVERTISED_HOST, '/graphql');
+
+const hitsQuery = `
+  query hitsQuery($filters: JSON, $first: Int) {
+    file {
+      hits(filters: $filters, first: $first) {
+        edges {
+          node {
+            donors {
+              hits {
+                edges {
+                  node {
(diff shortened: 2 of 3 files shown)


ciaranschutte:
investigating failing build in CI

src/resources/swagger/index.yaml
```diff
@@ -88,6 +88,35 @@ paths:
...
+        required: false
+        schema:
+          type: string
+    post:
+      tags:
+        - Clinical API
+      security:
+        - bearerAuth: []
+      summary: Download clinical data for all available files from filter.
+      description: Download clinical data for all available files from filter.
```
justincorrigible, line 106: thinking this may be a valid nuance?
```suggestion
      summary: Download clinical data for all available files from (optional) filters.
      description: Download clinical data for all available files from (optional) filters.
```

src/routes/clinical/arranger.ts
```diff
@@ -0,0 +1,95 @@
+import { ADVERTISED_HOST } from 'config';
```
justincorrigible, line 1: naive question: is `config` an npm package or an internal module with an absolute path? if the latter, this looks weird in the same set of imports as the rest.

src/routes/clinical/arranger.ts
```diff
@@ -0,0 +1,95 @@
...
+	egoJwt: string;
+}): Promise<string[]> {
+	const requestConfig = {
+		method: 'post',
+		headers: { Authorization: `Bearer ${egoJwt}`, 'Content-Type': 'application/json' },
+	};
+	try {
+		const queryClient = createQueryClient({ requestConfig });
+		const queryVariables = { filters: requestFilter };
+		// need two queries to pass "total" into hits filter to specify offset range
```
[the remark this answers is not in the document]
justincorrigible: ...for instance, line 76 does say "this gets the total number of hits" -> and absolute numerical value. clear. but `totalData` would be that "total hits" and what "total" else?

not sure I'm making sense... but the code + comment seem opaque/arcane to me, is ultimately my point.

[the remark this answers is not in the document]
justincorrigible: ok. rereading these lines a few more times, I think the issue I'm having is in how the code is separated into different lines; and that forces variable creation for every step, which in turn brings out how naming things is hard...

it occurs to me lines 75 - 76 could look more like the following (which I realize mixes then and await), but conceptually explains better the intent:

```
const totalHits = await queryClient({ 
  query: makeRequest({ query: totalQuery, variables: queryVariables })
}).then(response => get(response, 'data.file.hits.total'));
```

...maybe query client should also take a callback as an optional param?
"in case you want to do something with the response".

src/resources/swagger/index.yaml
justincorrigible: thinking further on this, the "all" part of this endpoint implies to me no filtering should be possible. I mean why would a query give you partial results, unless it is due to pagination or hard limits?

src/routes/clinical/clinical-api.ts
```diff
@@ -172,6 +160,63 @@ const createClincalApiRouter = async (esClient: Client, egoClient: EgoClient) =>
 			});
 		}
 	});
+
+	router.use('/donors/data-for-all-files', async (req: AuthenticatedRequest, res) => {
```
justincorrigible, line 164: if a filter is passed, this should be read as "'some' data about 'some' of the files".

I think "all" here is either redundant or misleading, unless it actually means "'some' data for all the files", which is not immediately explicit in the name

src/resources/swagger/index.yaml
joneubank: I would rephrase this a bit farther even:

```
Download clinical data associated with the donors for all files. A SQON filter can be provided to restrict the data retrieved to the files returned by a specific query.
```

src/routes/clinical/arranger.ts
```diff
@@ -0,0 +1,95 @@
...
+		} catch (error) {
+			console.error(`Failed to query server`, error);
+		}
+	};
+}
+
+async function queryArranger({
+	requestFilter,
+	egoJwt,
+}: {
```
joneubank, line 63: Powerful new exported function, should have TSDoc describing what it does, what inputs it requires and what it will return. Best to also include example of how to use it.

src/routes/clinical/clinical-api.ts
```diff
@@ -172,6 +160,63 @@ const createClincalApiRouter = async (esClient: Client, egoClient: EgoClient) =>
...
+
+			try {
+				// query GQL Arranger
+				// version 2.19 does not have direct access to Arranger apis directly
+				const uniqueDonors = await queryArranger({ requestFilter, egoJwt });
+
+				// Return all files in a zip with a manifest
+				const filename = `clinical_data_${Date.now()}`;
+
+				const response = await downloadDonorTsv(clinicalAuthClient, uniqueDonors);
```
joneubank, line 200: There are important filters that aren't being applied here. If you look at the `data-for-files` endpoint, we filter the files based on several important considerations:
- ✅  user permissions and file embargo state, these will be applied by arranger since you pass the user's auth information to the arranger request
- ❌ filtered by `has_clinical_data` is `true` - we should add this to the query we are making to arranger

joneubank (changes requested):
Almost all correct. I think we are missing a filter on our request to Arranger that filters the files by those that have clinical data (`has_clinical_data === true`). We should add an extra SQON filter to our arranger request to accomodate this:

```jsonc
{
  op: "and",
  content: [
    {
      op: "in",
      content: { fieldName: "has_clinical_data", value: [ "true" ] }
    },
    { /* user provided filter */ }
  ]
}
```

joneubank (approved):
Changes are good, need to remove test containers from any tests run in the `test` script in order to get this to run in Jenkins.

